PR #2, opened 2015-04-01
Add a custom completion rule for statements to find sponsor

davidZIVI:
- Add a custom completion rule for statements to find sponsor


ecino:
Completion does not work. Every line is matched by more than one sponsor, the comparison is maybe not restrictive enough.

PR #4, opened 2015-04-20
1.5 Change advance_billing_months to integer

davidZIVI:
Add functions to manage advance_billing_changes (in progress)
Add migrations script to move data from advance_billing to advance_billing_months


recurring_contract/__openerp__.py
```diff
@@ -30,7 +30,7 @@
 {
     'name': 'Recurring contract',
     'summary': 'Contract for recurring invoicing',
-    'version': '1.1',
+    'version': '1.4',
```
ecino: From 1.1 to 1.4 ? Either we increment normally (i.e. 1.2) or we follow the branch version of the whole project (i.e. 1.5).

recurring_contract/model/contract_group.py
```diff
@@ -82,22 +91,102 @@ def _get_groups_from_contract(self, cr, uid, ids, context=None):
...
         'recurring_value': 1,
+        'advance_billing_months': 1,
     }
 
+    def _get_contract_ids(self, cr, uid, ids, context=None):
+        contract_ids = list()
+
+        for contract_group_id in self.browse(cr, uid, ids, context):
+            for contract in contract_group_id.contract_ids:
+                if contract.id not in contract_ids:
```
ecino: Can the same contract be in several groups ? I don't think so...  I think this test is useless. As well the whole procedure could be reduced to :
```contract_ids.extend([contract.id for contract in contract_group.contract_ids])```

File: recurring_contract/model/contract_group.py [hunk 3]
```diff
@@ -82,22 +91,102 @@ def _get_groups_from_contract(self, cr, uid, ids, context=None):
...
+        return contract_ids
+
     def write(self, cr, uid, ids, vals, context=None):
+
+        recurring_contract_obj = self.pool.get('recurring.contract')
+        contract_ids = self._get_contract_ids(cr, uid, ids, context)
+
+        if ('advance_billing_months' in vals):
+            old_advance_billing_months = self.browse(
+                cr, uid, ids, context)[0].advance_billing_months
```
ecino: It's not a good practice to take into consideration only the first id. Make the method handle the case where several ids would be passed.

recurring_contract/model/contract_group.py
```diff
@@ -82,22 +91,102 @@ def _get_groups_from_contract(self, cr, uid, ids, context=None):
...
+        if ('recurring_value' in vals or
+                'recurring_unit' in vals):
+            self.button_generate_invoices(cr, uid, ids, context=context)
+
         res = super(contract_group, self).write(cr, uid, ids, vals, context)
-        if 'advance_billing' in vals:
-            self.generate_invoices(cr, uid, ids, context=context)
         return res
 
+    def _on_next_invoice_change(
```
ecino: Method is never called (don't know if it's in progress but I still write it in case).

davidZIVI: The method is in progress

recurring_contract/model/recurring_contract.py
```diff
@@ -53,6 +54,7 @@ def unlink(self, cr, uid, ids, context=None):
 
 
 class recurring_contract_line(orm.Model):
+
```
ecino: Remove empty line.

recurring_contract/model/contract_group.py
```diff
@@ -28,13 +28,36 @@ class contract_group(orm.Model):
     _inherit = 'mail.thread'
     _rec_name = 'ref'
 
+    def _get_change_methods(self, cr, uid, context=None):
+        ''' Method for invoice generation '''
```
ecino: Double quotes

recurring_contract/model/contract_group.py
```diff
@@ -82,27 +103,116 @@ def _get_groups_from_contract(self, cr, uid, ids, context=None):
...
-        if 'advance_billing' in vals:
-            self.generate_invoices(cr, uid, ids, context=context)
+        res = True
+
+        for group in self.browse(
+                cr, uid, ids, context):
+
+            if not group.next_invoice_date:
+                res = super(contract_group, self).write(
+                    cr, uid, group.id, vals, context) & res
```
ecino: replace & symbol by and

recurring_contract/model/contract_group.py
```diff
@@ -82,27 +103,116 @@ def _get_groups_from_contract(self, cr, uid, ids, context=None):
...
+                break
+
+            change_method = (group.change_method if 'change_method' not in vals
```
ecino: you can replace by ```change_method = vals.get('change_method') or group.change_method```

recurring_contract/model/contract_group.py
```diff
@@ -82,27 +103,116 @@ def _get_groups_from_contract(self, cr, uid, ids, context=None):
...
         'recurring_value': 1,
+        'advance_billing_months': 1,
+        'change_method': 'do_nothing',
     }
 
     def write(self, cr, uid, ids, vals, context=None):
-        res = super(contract_group, self).write(cr, uid, ids, vals, context)
-        if 'advance_billing' in vals:
-            self.generate_invoices(cr, uid, ids, context=context)
+        res = True
```
ecino: A comment method would be welcome.

recurring_contract/model/contract_group.py
```diff
@@ -82,27 +103,116 @@ def _get_groups_from_contract(self, cr, uid, ids, context=None):
...
+                res = super(contract_group, self).write(
+                    cr, uid, group.id, vals, context) & res
+                break
+
+            change_method = (group.change_method if 'change_method' not in vals
+                             else vals['change_method'])
+            change_method = getattr(self, change_method)
+            old_advance_billing_months = group.advance_billing_months
+
+            if ('advance_billing_months' in vals):
```
ecino: More comments would be welcome

recurring_contract/model/contract_group.py
```diff
@@ -82,27 +103,116 @@ def _get_groups_from_contract(self, cr, uid, ids, context=None):
...
+        if ('advance_billing_months' in vals or
+                'recurring_value' in vals or
+                'recurring_unit' in vals or
+                'change_method' in vals):
+            invoicer_id = self.generate_invoices(cr, uid, ids, context=context)
+            self.validate_invoices(cr, uid, invoicer_id, context)
+
         return res
 
+    """ Calculate the since_date to clean """
```
ecino: This comment is not well placed. Put it after method declaration.

recurring_contract/model/recurring_contract.py
```diff
@@ -191,13 +205,27 @@ def _check_unique_reference(self, cr, uid, ids, context=None):
...
             vals['reference'] = self.pool.get('ir.sequence').next_by_code(
                 cr, uid, 'recurring.contract.ref', context=context)
         return super(recurring_contract, self).create(cr, uid, vals,
                                                       context=context)
 
+    def write(self, cr, uid, ids, vals, context=None):
+        res = super(recurring_contract, self).write(
+            cr, uid, ids, vals, context=context)
+
+        """ Perform various checks when a contract is modified. """
```
ecino: Comment after method definition.

recurring_contract/model/recurring_contract.py
```diff
@@ -311,11 +360,81 @@ def _compute_next_invoice_date(self, contract):
...
+            self, cr, uid, ids, new_invoice_date, context=None):
+        for contract in self.browse(cr, uid, ids, context):
+            if contract.next_invoice_date:
+                next_invoice_date = datetime.strptime(
+                    contract.next_invoice_date, DF)
+                new_invoice_date = datetime.strptime(new_invoice_date, DF)
+                if (next_invoice_date > new_invoice_date):
+                    raise orm.except_orm(
+                        'Error',
+                        _('You cannot set the next invoice date'
```
ecino: Be more specific in error message by telling we cannot rewind next_invoice_date.

recurring_contract/model/recurring_contract.py
```diff
@@ -291,6 +326,20 @@ def clean_invoices(self, cr, uid, ids, context=None, since_date=None,
...
 
+        for contract in self.browse(cr, uid, ids, context):
+            next_invoice_date = since_date
+
+            if contract.last_paid_invoice_date:
+                next_invoice_date = max(
+                    datetime.strptime(
+                        contract.last_paid_invoice_date, DF), since_date)
+
+            self.write(
```
ecino: I think you should call super write method to avoid having an error when the next_invoice_date is rewinded.

ecino: I see two problematic use cases with this addition :

* Suspension : it will call clean_invoices to avoid paying in the period of suspension, but the next_invoice_date should not be updated !
* Contract termination : it will call clean_invoices but next_invoice_date does not not need to be updated.

You should avoid this either by adding a value in the context for preventing the modification of next_invoice_date, or adding a parameter to the method. You can then search all the calls to this method and see where it makes sense to update next_invoice_date and where it doesn't make sense to do it.

recurring_contract/model/recurring_contract.py
```diff
@@ -311,11 +360,81 @@ def _compute_next_invoice_date(self, contract):
...
+        for contract in self.browse(cr, uid, ids, context):
+            if contract.next_invoice_date:
+                next_invoice_date = datetime.strptime(
+                    contract.next_invoice_date, DF)
+                new_invoice_date = datetime.strptime(new_invoice_date, DF)
+                if (next_invoice_date > new_invoice_date):
+                    raise orm.except_orm(
+                        'Error',
+                        _('You cannot set the next invoice date'
+                          'at {}.'.format(new_invoice_date)))
```
ecino: {0}

PR #6, opened 2015-05-29
1.5

steveferry:
Following features added:
- Raise exception if sponsorship is not defined when new specifics products are added

### DIFF
--- a/recurring_contract/model/recurring_invoicer.py
+++ b/recurring_contract/model/recurring_invoicer.py
@@ -44,7 +44,6 @@ class recurring_invoicer(orm.Model):
             cr, uid, 'rec.invoicer.ident', context=context),
         'generation_date': lambda self, cr, uid, context:
         datetime.today().strftime(DF),
-        'source': 'default value',
     }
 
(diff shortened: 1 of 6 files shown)


recurring_contract/model/recurring_invoicer.py
```diff
@@ -43,6 +44,7 @@ class recurring_invoicer(orm.Model):
             cr, uid, 'rec.invoicer.ident', context=context),
         'generation_date': lambda self, cr, uid, context:
         datetime.today().strftime(DF),
+        'source': 'default value',
```
ecino: Remove the default value, so that it is empty in default case...

recurring_contract/view/recurring_invoicer_view.xml
```diff
@@ -14,6 +14,7 @@
 		<field name="arch" type="xml">
 			<tree string="Recurring invoicers" create="false">
 				<field name="identifier" />
+                <field name="source" />
```
ecino: Wrong indentation

ecino:
Please fix this PR and I will merge.

PR #8, opened 2015-06-04
8.0

steveferry:
Move 7.0 modules in __unported__


recurring_contract/model/contract_group.py
```diff
@@ -12,20 +12,21 @@
 from datetime import datetime
 from dateutil.relativedelta import relativedelta
 
-from openerp.osv import orm, fields
+import openerp
```
ecino: Is this import mandatory ?

recurring_contract/model/contract_group.py
```diff
@@ -34,89 +35,75 @@ def _get_change_methods(self, cr, uid, context=None):
...
+    def _get_last_paid_invoice(self):
         res = dict()
-        for group in self.browse(cr, uid, ids, context):
+        for group in self:
             res[group.id] = max([c.last_paid_invoice_date
                                  for c in group.contract_ids] or [False])
         return res
 
-    def _get_groups_from_contract(self, cr, uid, ids, context=None):
+    def _get_groups_from_contract(self):
```
ecino: This method is no longer needed.

recurring_contract/model/contract_group.py
```diff
@@ -125,194 +112,162 @@ def write(self, cr, uid, ids, vals, context=None):
...
+            self.validate_invoices(invoicer_id)
 
         return res
-
-    def button_generate_invoices(self, cr, uid, ids, context=None):
-        invoicer_id = self.generate_invoices(cr, uid, ids, context=context)
-        self.validate_invoices(cr, uid, invoicer_id, context)
+
+    def button_generate_invoices(self):
+        invoicer_id = self.generate_invoices()
```
ecino: Maybe we should rename `invoicer_id` to `invoicer`, as if I understand correctly, the returned value is a Record, not just an id.

recurring_contract/model/contract_group.py
```diff
@@ -34,89 +35,75 @@ def _get_change_methods(self, cr, uid, context=None):
...
-                                 for c in group.contract_ids
-                                 if c.state in self._get_gen_states()] or
-                                [False])
-        return res
+    @api.depends('contract_ids.next_invoice_date', 'contract_ids.state')
+    def _get_next_invoice_date(self):
+        next_inv_date = min(
+            [c.next_invoice_date for c in self.contract_ids
+             if c.state in self._get_gen_states()] or [False])
+        self.next_invoice_date = next_inv_date
```
ecino: I read in documentation that @api.multi is the default behaviour, so here self is a Recordset. Is it valid to do `self.next_invoice_date = next_inv_date` on a recordset or should it be `self.write({'next_invoice_date: next_inv_date})` ? If both are valid, would the last one trigger less queries on the database ?

recurring_contract/model/contract_group.py
```diff
@@ -34,89 +35,75 @@ def _get_change_methods(self, cr, uid, context=None):
...
 
-    def _get_last_paid_invoice(self, cr, uid, ids, name, args, context=None):
+    def _get_last_paid_invoice(self):
         res = dict()
-        for group in self.browse(cr, uid, ids, context):
+        for group in self:
```
ecino: I think here instead of returning res you should assign the result in each iteration doing `group.last_paid_invoice_date = max(...)`

recurring_contract/model/invoice.py
```diff
@@ -9,58 +9,59 @@
...
+        'recurring.invoicer', _('Invoicer'))
 
 
-class account_invoice_line(orm.Model):
+class account_invoice_line(models.Model):
     _name = 'account.invoice.line'
     _inherit = 'account.invoice.line'
 
-    def _get_dates(self, cr, uid, ids, name, args, context=None):
+    def _get_dates(self):
```
ecino: This method is no longer needed.

recurring_contract/model/invoice.py
```diff
@@ -9,58 +9,59 @@
...
         res = {}
-        for line in self.browse(cr, uid, ids, context):
+        for line in self:
             res[line.id] = line.invoice_id.date_due or False
 
         return res
 
-    def _get_states(self, cr, uid, ids, name, args, context=None):
+    def _get_states(self):
```
ecino: No longer needed.

recurring_contract/model/invoice.py
```diff
@@ -9,58 +9,59 @@
...
-            _get_states, string=_('State'), readonly=True, type='selection',
-            selection=[('draft', 'Draft'),
-                       ('proforma', 'Pro-forma'),
-                       ('proforma2', 'Pro-forma'),
-                       ('open', 'Open'),
-                       ('paid', 'Paid'),
-                       ('cancel', 'Cancelled')],
-            store={'account.invoice': (_get_invoice_lines, ['state'], 20)}),
-    }
+    @api.depends('due_date')
```
ecino: I think the depends should be on `invoice_id.date_due`

recurring_contract/model/recurring_contract.py
```diff
@@ -81,146 +61,123 @@ class recurring_contract(orm.Model):
...
-        for self_contract in self.browse(cr, uid, ids, context=context):
-            if self_contract.reference and self_contract.reference in lst:
-                return False
+    @api.depends('contract_line_ids', 'contract_line_ids.amount',
+                 'contract_line_ids.quantity')
+    def _get_total_amount(self):
+        self.total_amount = sum([line.subtotal for line in
+                                 self.contract_line_ids])
+
+    # def _get_contract_from_group(group_obj, cr, uid, group_ids, context=None):
```
ecino: You can remove the commented code.

ecino:
:+1:

PR #9, opened 2015-06-08
1.5 review

ecino:
# 1.5 Improvements

Contains new features added after split of contracts.

## advanced_reconcile

- Change attribution wizard : utility for directly modifying a paid
  invoice which will unreconcile / cancel / modify / validate / pay
  the selected invoices / payments.
- Add a comment on splitted payments move lines

## analytic_attribution

Instead of using analytic distributions, we will now use attributions for
creating custom rules that will dispatch analytic lines into several analytic
accounts. We therefore won't need distributions anymore and need only to provide
an analytic account for invoices / move lines / ...

## completion

- New rules for LSV / DD allowing to look into cancelled and paid invoices.
- Various bug fixes and improvements

## recurring_contract

- Change of fields which were ambiguous by their name and types (advance_billing, frequency, ...)
- Be more generic
- Change methods for being able to add methods called when contracts/groups are modified


account_advanced_reconcile_compassion/wizard/reconcile_fund_wizard.py
```diff
@@ -125,12 +132,12 @@ def reconcile_with_fund(self, cr, uid, ids, context=None):
...
 
     def _generate_invoice_line(self, cr, uid, invoice_id, product, price,
-                               partner_id, context=None):
+                               partner_id, contract_ids, context=None):
 
         inv_line_data = {
             'name': product.name,
             'account_id': product.property_account_income.id,
-            'price_unit': price,
+            'price_unit': price / len(contract_ids),
```
CSester: As you divide the price by nb of contract to get the unit price, shouldn't the quantity be the nb of contract to get the price of the invoice line be the actual price given to the function ?

account_advanced_reconcile_compassion/wizard/reconcile_fund_wizard.py
```diff
@@ -148,6 +155,8 @@ def _generate_invoice_line(self, cr, uid, invoice_id, product, price,
 
         res['name'] = product.name
 
-        self.pool.get('account.invoice.line').create(
-            cr, uid, inv_line_data, context=context)
+        for contract_id in contract_ids:
```
CSester: Ok, ignore my last comment

account_statement_completion_compassion/model/completion_rules.py
```diff
@@ -263,14 +255,14 @@ def _generate_invoice(self, cr, uid, st_line, partner, context=None):
...
-        if product.name not in GIFT_TYPES:
+        if product.product_tmpl_id.categ_id.name != 'Sponsor gifts':
             # Validate the invoice
             wf_service = netsvc.LocalService('workflow')
             wf_service.trg_validate(
                 uid, 'account.invoice', invoice_id, 'invoice_open', cr)
 
         # Birthday Gift
-        elif product.name == GIFT_TYPES[0]:
+        elif product.name == 'Birthday Gift':
```
CSester: Why do you prefer usage of hardcoded text rather than usage of constants ?

account_statement_completion_compassion/model/completion_rules.py
```diff
@@ -358,3 +359,67 @@ def _generate_invoice_line(self, cr, uid, invoice_id, product, st_line,
...
+                partner = invoice_obj.browse(
+                    cr, uid, invoice_ids, context=context)[0].partner_id
+        return partner
+
+    def get_sponsor_name(self, cr, uid, st_line, context=None):
+        res = dict()
+        label = st_line['label']
+
+        # Fix encoding problem
+        reload(sys)
```
CSester: Why reload? Reload is made to reload the module if it changed, but I didn't (and hopefully will never) see any changes in the sys module.

recurring_contract/migrations/1.2/post-migration.py
```diff
@@ -0,0 +1,39 @@
...
+#    @author: Coninckx David <david@coninckx.com>
+#
+#    The licence is in the file __openerp__.py
+#
+##############################################################################
+import sys
+
+
+def migrate(cr, version):
+    reload(sys)
```
CSester: Same question as before

recurring_contract/migrations/1.2/post-migration.py
```diff
@@ -0,0 +1,39 @@
...
+    )
+
+    contract_groups = cr.fetchall()
+
+    for contract_group in contract_groups:
+        delay = delay_dict[contract_group[1]] or 1
+        cr.execute(
+            '''
+            UPDATE recurring_contract_group
+            SET recurring_value = {0}, advance_billing_months = {0}
```
CSester: I didn't remember exactly what each field contains:
 - Recurring value: how many month are grouped on an invoice
 - advance_billing_months: ?

account_advanced_reconcile_compassion/wizard/reconcile_fund_wizard.py
ecino: Which one ?

account_statement_completion_compassion/model/completion_rules.py
ecino: It's a trick to change encoding to UTF8 and avoid some errors with accented characters.

http://stackoverflow.com/questions/2276200/changing-default-encoding-of-python

http://stackoverflow.com/questions/3828723/why-we-need-sys-setdefaultencodingutf-8-in-a-py-script

recurring_contract/migrations/1.2/post-migration.py
ecino: Thank you, you pointed out a potential huge problem if we would have kept this ! Indeed, we don't wan't to update the recurring_value. However we maybe do it manually for non-sponsorship contracts so that we stop generating invoices each month if the partner does not pay each month.

account_advanced_reconcile_compassion/wizard/reconcile_fund_wizard.py
CSester: The one about dividing by number of contracts

ecino:
See #10 for the corrections.

PR #10, opened 2015-06-17
1.5 Additions and corrections

ecino:
- Albert did some tests and corrected some code based on the results
- Corrections for the last review
- Travis should be successful


ecino:
I think we won't need this review anymore, as it becomes old and we are now migrating all code to V8, which simplify and changes a lot the code. So I close it.

PR #16, opened 2015-08-06
Modification of demo of recurring_contract

albertsh:
- Add of accounts 
- Change of type of accounts
- Add of account_type2


recurring_contract/model/contract_group.py
```diff
@@ -307,7 +305,7 @@ def _setup_inv_line_data(self, cr, uid, contract_line, invoice_id,
         return inv_line_data
 
     def _generate_invoice_lines(self, cr, uid, contract, invoice_id,
-                                context=None):
+                                context=None):
```
ecino: Remove blank spaces

ecino:
The test for this repository are ok. We will see why we have tests of other repository... :+1:

PR #24, opened 2015-08-31
add of language for each partner

albertsh:


recurring_contract/tests/test_base_contract.py
```diff
@@ -31,7 +31,8 @@ def setUp(self):
         property_account_payable = account_obj.search([
             ('type', '=', 'payable'),
             ('user_type', '=', account_type)]).ids[0]
-
+        category = self.env['res.partner.category'].create({
+            'name': 'sponsor'})
```
ecino: This should not be in `test_base_contract` as there are no sponsors in this module. Instead, you could move this create in the setup of test of module `contract_compassion` and then: 
```
self.partners.write({'category_id': [(4, category.id)]})
````

File: recurring_contract/tests/test_base_contract.py [hunk 2]
```diff
@@ -56,7 +57,16 @@ def setUp(self):
             'property_account_payable': property_account_payable,
             'notify_email': 'always',
             'ref': '00003333',
+            'lang': 'en_US',
         })
+        self.partners += partner_obj.create(
+            {
+                'lang': 'en_US',
+                'name': 'Client 37',
```
ecino: Why did you add another partner ? Plus this one has no reference.

PR #26, opened 2015-11-13
fix color options in dropdown of sponsorship field in bank statement reconcile view

rz-comp:
the options are now passed into the `init()` function of the field, as there was no other way to add an `options` attribute directly to the field object otherwise


ecino:
:+1:

PR #27, opened 2015-11-18
fix invoice generation job

rz-comp:


ecino:
Thank you, I will integrate those changes in my next commit (with test fixes)

PR #47, opened 2018-01-18
CO-1365 - Birthday gif date is no longer linked to invoice date

cmtCousteau:


recurring_contract/models/contract_group.py
```diff
@@ -153,7 +153,7 @@ def generate_invoices(self, invoicer=None):
                 delay += relativedelta(minutes=1)
             self.with_delay(eta=delay)._generate_invoices(invoicer)
         else:
-            self._generate_invoices(invoicer)
+            self.with_context(lsv=True)._generate_invoices(invoicer)
```
ecino, line 156: This change should not be here. You should instead inherit this method in module `sponsorship_compassion` and call the super with the new context. I will close your pull request and you can reset your compassion-accounting repository :
```
git fetch compassion
git reset --hard compassion/10.0
git push -f
```

PR #59, opened 2018-07-13
Co 870 analytic distribution

nbornand:


account_analytic_attribution/models/account_analytic_attribution.py
```diff
@@ -40,14 +40,6 @@ class AccountAttribution(models.Model):
     sequence = fields.Integer()
     next_fiscal_year = fields.Date(compute='_compute_next_fy')
 
-    @api.model
-    def next_fiscal_year(self):
```
ecino, line 44: Was this method never used?

nbornand: I couldn't find any usage

ecino (approved):

PR #77, opened 2019-04-04
CO2468

vwvw:
Improved reconcile function and fixed names breaking CI

PR #79, opened 2019-04-09
CO2468 (#77)

ecino:
* Fix for CO-2468. addtl_ntry_inf was added to the name.

* Removed numerours calls to search

* Disabled reconciliation for too big sets of lines resulting in the
script never finishing.

pylint

* Address @ecino comment


ecino:
Tests fails are on OCA side.

PR #83, opened 2019-08-13
Migrate recurring_contract

ecino:
Replaces #82


recurring_contract/__init__.py
```diff
@@ -1,13 +1,2 @@
...
-#    @author: Cyril Sester <csester@compassion.ch>
-#
-#    The licence is in the file __manifest__.py
-#
-##############################################################################
-
-from . import models
-from . import wizard
+from . import models
+from . import wizards
```
eicher31:
```suggestion
from . import wizards

```

recurring_contract/data/contract_expire_cron.xml
```diff
@@ -1,22 +1,20 @@
...
+    <data noupdate="1">
+        <record id="recurring_contract_expire_cron" model="ir.cron">
+            <field name="name">Check end of contract</field>
+            <field name="active" eval="True"/>
+            <field name="user_id" ref="base.user_root"/>
+            <field name="interval_number">1</field>
+            <field name="interval_type">days</field>
+            <field name="numbercall">-1</field>
+            <field name="model_id" ref="model_recurring_contract"/>
+            <field name="function" eval="'end_date_reached'" />
```
eicher31: Why use double quote and quote together?

eicher31 (approved):

recurring_contract/data/contract_expire_cron.xml
ecino: Because of eval which executes python code and here we want a string. Best would be to remove eval and put the string as a normal xml tag.
```suggestion
            <field name="function">end_date_reached</field>
```

PR #86, opened 2019-09-10
account_analytic_attribution migration

quentingigon:


ecino (changes requested):
Please review and correct so that the Travis build is good for both Lint and Tests. It may be necessary to adapt and add unit tests as well.

PR #103, opened 2020-02-18
CO-3051 add activity to recurring_contracts

josueBulle:
related:
https://github.com/CompassionCH/compassion-modules/pull/1293
https://github.com/CompassionCH/compassion-switzerland/pull/1016

### DIFF
--- a/recurring_contract/views/recurring_contract_view.xml
+++ b/recurring_contract/views/recurring_contract_view.xml
@@ -97,6 +97,7 @@
                 </sheet>
                 <div class="oe_chatter">
                     <field name="message_follower_ids" widget="mail_followers"/>
+                    <field name="activity_ids" widget="mail_activity"/>
                     <field name="message_ids" widget="mail_thread"/>
                 </div>
             </form>


ecino:
Can you add OCA repository for mis_builder module so that travis goes green please?

mis_builder_spn_event_info/__manifest__.py
```diff
@@ -18,6 +18,6 @@
         'security/mis_spn_event_info.xml',
         'views/mis_spn_event_info.xml',
     ],
-    'installable': False,
+    'installable': True,
```
ecino: This should not be present in this PR

ecino (changes requested):

ecino:
After reflextion, can you move modules mis_builder* into compassion-switzerland repository ? It makes more sense there I think.

ecino (changes requested):
I think you removed the class inheritance by mistake.

josueBulle:
The model `recurring_contract` was already inheriting `mail.activity.mixin`

ecino (approved):

PR #108, opened 2020-05-12
CO-3158 Fix product_id field of reconcile js view

SylvainLosey:
The product field now seems to work properly, however I don't exactly know how the module is supposed to work so this might need further testing.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ addons:
 
 env:
   global:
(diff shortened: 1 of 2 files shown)


ecino (approved):
:+1: Should I wait your task CO-3190 before merging ?

SylvainLosey:
Yes I think so, everything will be in CO-3190

SylvainLosey:
Replaced by CO-3190, PR is in compassion-switzerland.

PR #109, opened 2020-05-12
L10n ch import camt054 mig v12

quentingigon:
To test this, I ran the CRON manually with this value: model.camt054_reconcile("1098",)
Previously, I had to reconstruct some data in my v12 database, as values for _acct_svcr_ref_ were lost during a migration. For that I used those psql commands:

- `CREATE TEMP TABLE tmp_lines (id int, acct_svcr_ref text);`
- `COPY tmp_lines FROM '/home/qgigon/account_move_lines_10.csv' (FORMAT csv);`
- `UPDATE account_move_line SET acct_svcr_ref = tmp_lines.acct_svcr_ref FROM tmp_lines WHERE account_move_line.id = tmp_lines.id;`
- `DROP TABLE tmp_lines;`

The CSV file used was created with pgadmin from a v10 database and contained _id_ +  _acct_svcr_ref_ for all rows with _acct_svcr_ref_ != ''

The CRON finished in around 25 minutes so I think this is an improvement compared to before but it should be tested on the server to be sure.


quentingigon:
Travis fails because it doesn't find the module _account_bank_statement_import_camt_details_ as it is in bank-statement-import

ecino:
You can add these lines into `oca_dependenciest.txt`
```
bank-statement-import https://github.com/CompassionCH/bank-statement-import 12.0-camt-details
l10n-switzerland https://github.com/CompassionCH/l10n-switzerland 12.0-everything-migrated
```

l10n_ch_import_camt054/__manifest__.py
```diff
@@ -8,13 +8,13 @@
     'depends': [
         'account_bank_statement_import_camt_details',
         'l10n_ch_fds_postfinance',
-        'account_payment_line_cancel',
+        # 'account_payment_line_cancel',
```
ecino: You can remove completely

l10n_ch_import_camt054/demo/test_data.yml
```diff
@@ -27,7 +27,7 @@
 # Create account
 #-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=
 -
-  !record {model: account.account, id: post_account, view: False}:
+  !record {model: account.account, id: post_account, view: False}:super
```
ecino: `yaml` files are no more supported! You should replace it with `xml` file.

ecino (changes requested):

ecino:
Line 95 in custom parser won't be working (https://github.com/CompassionCH/compassion-accounting/blob/9f61c3f636aca78d900ec949d20c7a5853ed5e0a/l10n_ch_import_camt054/models/custom_parser.py#L95) as it uses the module `payment_line_cancel` which was removed and replaced by `account_payment_return`. You can take inspiration from module `l10n_ch_payment_return_sepa` which is importing files and doing as well the cancellation by creating a `account.payment.return` object : 
https://github.com/CompassionCH/l10n-switzerland/blob/12.0-everything-migrated/l10n_ch_payment_return_sepa/wizards/payment_return_import.py#L65
@davidwul Could you check that what I say is correct ?

quentingigon:
I am waiting for the confirmation from David that the proposed change is correct

ecino:
It looks like `Crypto` must be added in the requirements.txt

quentingigon:
Travis does not start

ecino:
Please fix the tests which are still failing.

PR #112, opened 2020-06-09
CO-3248 12.0 add analytic tags

wvmrtn:
CO-3248 add analytic tag to account.asset with views


account_analytic_compassion/__manifest__.py
```diff
@@ -34,12 +34,15 @@
     'license': 'AGPL-3',
     'website': 'http://www.compassion.ch',
     'depends': [
+        'account_analytic_attribution',
```
ecino: account_analytic_attribution is a module that we want to remove. Why did you need to add this dependency? I think it should be avoided.

account_analytic_compassion/models/account_asset.py
```diff
@@ -0,0 +1,53 @@
...
+from odoo import models, fields
+
+
+class AccountAsset(models.Model):
+    _inherit = 'account.asset'
+
+    # Many2many field to account.analytic.tag table
+    analytic_tag_ids = fields.Many2many(
+        comodel_name='account.analytic.tag',
+        string='Analytic Tag',
```
ecino:
```suggestion
        string='Analytic Tags',
```

account_analytic_compassion/models/account_asset.py
```diff
@@ -0,0 +1,53 @@
...
+            'credit': credit,
+            'debit': debit,
+            'journal_id': asset.profile_id.journal_id.id,
+            'partner_id': asset.partner_id.id,
+            'analytic_account_id': analytic_id,
+            'date': depreciation_date,
+            'asset_id': asset.id,
+            'analytic_tag_ids': asset.analytic_tag_ids.ids
+        }
+        return move_line_data
```
ecino: It is not a good practice to completely rewrite a method, instead of making usage of the super call. The parent method returns a dictionary. You can use super to have the result, modify the result in your method, and then return the final result, like this:
```suggestion
        move_line_data = super()._setup_move_line_data(
            depreciation_date, account, ml_type, move)
        if ml_type == 'expense':
            move_line_data["analytic_tag_ids"] = [(6, 0, asset.analytic_tag_ids.ids)]
        return move_line_data
```
[(6, 0, asset.analytic_tag_ids.ids)] is a special command for writing in many2many field: see https://www.odoo.com/documentation/12.0/reference/orm.html#odoo.models.Model.write for reference

ecino (changes requested):

account_analytic_compassion/models/account_asset.py
wvmrtn: Tried:
```python
 move_line_data["analytic_tag_ids"] = [(6, 0, asset.analytic_tag_ids.ids)]
```
And Travis tests did not pass.
Changed to:
```python
move_line_data["analytic_tag_ids"] = asset.analytic_tag_ids.ids
```

ecino (approved):

PR #116, opened 2020-06-24
CO-3285 Add ambassador field to sponsorship

SylvainLosey:


ecino:
user_id doesn't exist in this module. The change should be done in `sponsorship_compassion` module.

PR #127, opened 2021-05-04
Refactor contract clean

JonathanGuerne:
- does not cancel paid invoice on clean by default
- add context value to prevent automatic cleaning on next_invoice_date change if needed
- improve invoice generation to reopen newly cancelled invoices rather than creating new ones
- add test to assess good behavior

### DIFF
--- a/recurring_contract/models/contract_group.py
+++ b/recurring_contract/models/contract_group.py
@@ -134,7 +134,7 @@ class ContractGroup(models.Model):
         pass
 
     @api.multi
-    def generate_invoices(self, invoicer=None):
+    def generate_invoices(self, invoicer=None, cancelled_invoices=None):
         """ By default, launch asynchronous job to perform the task.
             Context value async_mode set to False can force to perform
             the task immediately.
@@ -149,9 +149,9 @@ class ContractGroup(models.Model):
             delay = datetime.today()
             if jobs:
                 delay += relativedelta(minutes=1)
-            self.with_delay(eta=delay)._generate_invoices(invoicer)
+            self.with_delay(eta=delay)._generate_invoices(invoicer, cancelled_invoices=cancelled_invoices)
         else:
-            self._generate_invoices(invoicer)
+            self._generate_invoices(invoicer, cancelled_invoices=cancelled_invoices)
         return invoicer
 
     @api.multi
@@ -179,13 +179,15 @@ class ContractGroup(models.Model):
     @api.multi
     @job(default_channel='root.recurring_invoicer')
     @related_action(action='related_action_invoicer')
-    def _generate_invoices(self, invoicer=None):
+    def _generate_invoices(self, invoicer=None, cancelled_invoices=None):
         """ Checks all contracts and generate invoices if needed.
         Create an invoice per contract group per date.
         """
         logger.info("Invoice generation started.")
         if invoicer is None:
             invoicer = self.env['recurring.invoicer'].create({})
+        if cancelled_invoices is None:
+            cancelled_invoices = self.env["account.invoice"]
         inv_obj = self.env['account.invoice']
         gen_states = self._get_gen_states()
         journal = self.env['account.journal'].search([
@@ -216,9 +218,20 @@ class ContractGroup(models.Model):
                 if not contracts:
                     break
                 try:
+                    inv_to_reopen = cancelled_invoices.filtered(lambda inv: inv.date_invoice == current_date)
+
                     inv_data = contract_group._setup_inv_data(
                         journal, invoicer, contracts)
-                    invoice = inv_obj.create(inv_data)
+                    if not inv_to_reopen:
+                        invoice = inv_obj.create(inv_data)
+                    else:
+                        inv_to_reopen.action_invoice_draft()
+                        inv_to_reopen.env.clear()
+                        old_lines = inv_to_reopen.invoice_line_ids.filtered(
+                            lambda line: line.contract_id.id in contracts.ids)
+                        old_lines.unlink()
+                        inv_to_reopen.write(inv_data)
+                        invoice = inv_to_reopen
                     if invoice.invoice_line_ids:
                         invoice.action_invoice_open()
                     else:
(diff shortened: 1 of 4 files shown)


recurring_contract/models/contract_group.py
```diff
@@ -247,14 +260,20 @@ def _clean_generate_invoices(self):
         contract group.
         """
         res = self.env['account.invoice']
+        since_date = date.today()
         for group in self:
-            since_date = date.today()
-            if group.last_paid_invoice_date:
-                last_paid_invoice_date = group.last_paid_invoice_date
-                since_date = max(since_date, last_paid_invoice_date)
```
ecino, line 254: Can you please explain why you removed this code, are we sure it's not needed?

recurring_contract/models/recurring_contract.py
```diff
@@ -144,11 +144,14 @@ def create(self, vals):
...
     def write(self, vals):
         """ Perform various checks when a contract is modified. """
-        if vals.get('next_invoice_date'):
+        if vals.get('next_invoice_date') and not self.env.context.get("no_clean_on_write", False):
             self._on_change_next_invoice_date(vals['next_invoice_date'])
 
         res = super().write(vals)
 
+        if "partner_id" in vals:
+            self.group_id.partner_id = vals["partner_id"]
```
ecino:
```suggestion
            self.mapped("group_id").write({"partner_id": vals["partner_id"]})
```

recurring_contract/models/recurring_contract.py
```diff
@@ -182,39 +185,54 @@ def unlink(self):
...
+
+        res = self.env["account.invoice"]
+
         for contract in self:
             if contract.state in gen_states:
-                last_invoice_date = max([
-                    line.invoice_id.date_invoice for
-                    line in contract.invoice_line_ids
-                    if line.state in ('open', 'paid')] or [False])
+                last_invoice_date, is_paid = min(
```
ecino: Why are you taking the next future invoice instead of looking for the last invoice like it was previously done?
If that change is required, then I would suggest to rename the variable `last_invoice_date` to something more explanatory.

recurring_contract/models/recurring_contract.py
```diff
@@ -182,39 +185,54 @@ def unlink(self):
...
+                    [(line.invoice_id.date_invoice, line.state == "paid") for
+                     line in contract.invoice_line_ids
+                     if line.state in ("open", "paid") and
+                     line.invoice_id.date_invoice >= date.today()] or [
+                        (False, False)])
+
+                if last_invoice_date and is_paid:
+                    last_invoice_date += contract.group_id.get_relative_delta()
```
ecino: What if there are 3 future invoices in june, july and august? Your code would take june as `last_invoice_date` and then use july if june is paid. But July may also be paid. That might be an issue.

recurring_contract/models/recurring_contract.py
```diff
@@ -463,6 +480,7 @@ def _clean_invoices(self, since_date=None, to_date=None, keep_lines=None,
             invoice = inv_line.invoice_id
             # Check if invoice is empty after removing the invoice_lines
             # of the given contract
+            # TODO add check if call is made trough contract group
```
ecino: Is this something to address before we merge?

ecino (changes requested):

recurring_contract/models/contract_group.py
JonathanGuerne: the if wasn't used in the last version of the code. (neither since date nor last paid invoice are used) 

rewind_next_invoice_date handle the logic to regenerate invoices from today.

recurring_contract/models/recurring_contract.py
JonathanGuerne: We decided, I remember, to regenerate invoice with date >= to today. I first though about regenerating all opened invoices but you explained to me that this could lead to some issue. 

With this new regeneration approach multiple open invoices could be changed if they are still open.

I do agree that the name of the variable could be different.

recurring_contract/models/recurring_contract.py
JonathanGuerne: True. I didn't see this issue

recurring_contract/models/recurring_contract.py
ecino: Yes I now remember this discussion. If you find a better name I would be happy. Not mandatory change though. Maybe you could also rework the comment of the method for better explaining what it's doing.

recurring_contract/models/recurring_contract.py
```diff
@@ -307,7 +325,6 @@ def contract_waiting(self):
             raise UserError(_("Active contract cannot be put to waiting"))
         if self.filtered(lambda c: not c.total_amount):
             raise UserError(_("Please configure contract lines"))
-        self.mapped("group_id").generate_invoices()
```
ecino, line 310: Was it a duplicate call? Will the invoices still be generated when we validate a contract even after this change?

JonathanGuerne: It was not a duplicate call. This line cause issue on most of the test_case in the accounting module as they all make direct call to generate_invoice after the sponsorship is set to the 'waiting' state. 

No, they would have to be generated trough the button.

It would require a few changes in the test case but it should not be a mandatory thing to have this line remove if we want them to work properly.

ecino: Then please restore it as we want that the invoices automatically get generated when we validate the sponsorship. The test cases should be adapted to this behaviour if needed.

recurring_contract/models/recurring_contract.py
JonathanGuerne: I'll add a testcase for this scenario and see what changes need to be made

recurring_contract/models/recurring_contract.py
JonathanGuerne: Actually I was wrong... 

Looking back at the code in accounting I realized that it is not working properly.
The generation appends before the change of state. But invoices are only generated for waiting and active contract.

It was working because, as you suggested, this code was actually duplicated in an other module : https://github.com/CompassionCH/compassion-modules/blob/12.0/sponsorship_compassion/models/contracts.py#L1064 

Should  I left the code as is or do the opposite and handle the generation in accounting ?

ecino: You can leave it as it is now. Thanks for checking! :+1:

ecino:
Thanks for the changes, it looks fine now :+1:

ecino (approved):

JonathanGuerne:
I saw that travis was unable to pass all tests. It made me realized that I introduced a bug on my latest changes.

Handling invoice cleaning for a group of contract was already working. 
Line are removed one by one until only one remain then the invoice is left cancel.
Later on, in `_generate_invoices` all line are again generated.

PR #130, opened 2021-07-13
CO-3694 generate invoice based on advanced billing month and not always for one month in advance

JonathanGuerne:
use a memoize decorator to speed up the computation time.

### DIFF
--- a/recurring_contract/wizards/recurring_invoicer_wizard.py
+++ b/recurring_contract/wizards/recurring_invoicer_wizard.py
@@ -9,22 +9,6 @@
 ##############################################################################
 
 from odoo import fields, models, api
-import datetime
-from dateutil.relativedelta import relativedelta
-
-
-def memoize(func):
-    """The memoize decorator is used to speed up computation time
-    when a function is called multiple times with the same 'key'"""
-    memory = {}
-
-    def inner(key):
-        if key not in memory:
-            memory[key] = func(key)
-        return memory[key]
-
-    return inner
-
 
 class InvoicerWizard(models.TransientModel):
     ''' This wizard generate invoices from contract groups when launched.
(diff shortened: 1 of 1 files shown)


recurring_contract/wizards/recurring_invoicer_wizard.py
```diff
@@ -25,12 +37,16 @@ class InvoicerWizard(models.TransientModel):
 
     @api.multi
     def generate(self):
-        date_limit = datetime.date.today() + relativedelta(months=+1)
+
+        @memoize
```
ecino: Does this mean that if we don't restart the server it will always return the same limit date (because date.today() is no more computed) ?

recurring_contract/wizards/recurring_invoicer_wizard.py
```diff
@@ -25,12 +37,16 @@ class InvoicerWizard(models.TransientModel):
...
+        def get_limit_date(advance_months):
+            return datetime.date.today() + relativedelta(months=+advance_months)
 
         recurring_invoicer_obj = self.env['recurring.invoicer']
         contract_groups = self.env['recurring.contract.group'].search([
-            ('next_invoice_date', '<=', date_limit),
             ('next_invoice_date', '!=', False)])
+        contract_groups = contract_groups.filtered(
```
ecino: Why can't you continue to use the search instead of filtered? The search query would be much faster than the filtered function so I would suggest to revert that.

JonathanGuerne: this would require comparing a field to an other one 
```python
(['field_a', '<=', fun(field_b)])
```
I recall already have had to switch to a filter on this type of scenario. Is there an alternative ?

recurring_contract/wizards/recurring_invoicer_wizard.py
JonathanGuerne: You're right I hadn't thought about it. We could pass the day as well so it would only be computed once per day but it seems at this point like an unnecessary optimization.

ecino: So I think it's best to remove the optimization. I didn't know about this @memoize decorator, it could be useful for other things but here I think it's not.

recurring_contract/wizards/recurring_invoicer_wizard.py
ecino: Oh I didn't notice you are comparing two fields of the model. You are right that there is no way to do it directly in the search. I'm a bit afraid that we will have a massive performance drop with this change because we are loading almost all groups and then filtering on them. I would be more happy if the search could filter more in front before doing the filter in memory. But I don't have any solution to propose so if you don't find one either, let's move one like that.

ecino: Maybe the solution would be to execute a SQL query for fetching the ids and then use the browse method. 
```
self.env.cr.execute("""
    SELECT id FROM recurring_contract_group
    WHERE ...
"""")
gids = self.env.cr.fetchall()
contract_goups = self.browse(gids)
```

ecino (approved):

PR #135, opened 2021-12-21
CO-3851 the note in the bank statement should appear in the reconcili…

TheRaphael0000:
…ation line of the matched invoice

add a computed field on account.invoice that fetch the notes from the bank statements
show this field in the invoice view under the detailed information menu

PR #137, opened 2021-12-22
CO-3851 the note in the bank statement should appear in the reconciliation line of the matched invoice

TheRaphael0000:
use the message logger instead of a computed field to show the bank statements


TheRaphael0000:
suite of https://github.com/CompassionCH/compassion-accounting/pull/135

PR #156, opened 2022-11-18
CP-226 Update invoices infos 

simgonzalez:
When we update the payment method the current open invoice should be updated too

### DIFF
--- a/recurring_contract/models/contract_group.py
+++ b/recurring_contract/models/contract_group.py
@@ -225,16 +225,16 @@ class ContractGroup(models.Model):
                             f'failed during invoice generation',
                             exc_info=True)
                         break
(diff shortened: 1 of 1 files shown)


ecino (changes requested):
I think this change should rather be done in file recurring_contract.py:L600 inside the method `_update_invoice_lines` (and check there if it doesn't work, why).

simgonzalez:
@ecino Hello, yes but no, I agree that the modification isn't applied at the best place as it implies a new behaviour in something that's supposed to generate invoice and we add a simple update. Although, The method you pointed at isn't called in the proccess of updating the payment method (contract_group). Moreover, it's not that crazy to update invoices informations in a method that is supposed to create them.

ecino (commented):
I temporarily merge it to let David process the payments today. We can revert this once the change is properly made.

PR #157, opened 2022-11-21
CP-226 cleaner solution to adapt invoices data on payment method change

simgonzalez:


recurring_contract/models/recurring_contract.py
```diff
@@ -303,28 +303,20 @@ def rewind_next_invoice_date(self):
...
-                    if latest_paid_invoice_date else earliest_open_invoice_date
-
-                if rewind_invoice_date:
-                    res |= contract._clean_invoices(rewind_invoice_date)
-                    contract.with_context(no_clean_on_write=True).write({
-                        "next_invoice_date": rewind_invoice_date
-                    })
-                else:
+                rewind_invoice_date = earliest_open_invoice_date
+                if not rewind_invoice_date:
```
ecino: You completely removed the use of `latest_paid_invoice_date` which I don't think is correct. I would reintroduce it and use it first to try to determine the `rewind_invoice_date`.
```suggestion
                if latest_paid_invoice_date:
                    rewind_invoice_date = latest_paid_invoice_date + contract.group_id.get_relative_delta()
                else:
                    rewind_invoice_date = earliest_open_invoice_date
                if not rewind_invoice_date:
```

ecino (changes requested):

ecino (approved):
Can this replace the temporary patch in production, as it is?

PR #159, opened 2022-12-02
14.0 refactor invoices cp-265

simgonzalez:


recurring_contract/models/move.py
```diff
@@ -178,8 +177,101 @@ def find_sum(numbers, target, partial=None):
...
+        self.ensure_one()
+        # Build the dictionnary
+        inv_val_dict = dict()
+        inv_val_dict[self.name] = {
+            'invoice_line_ids': self._build_invoice_lines(inv_lines) if inv_lines else self.invoice_line_ids.ids,
+            'date': due_date if due_date else self.date,
+            'payment_reference': ref if ref else self.payment_reference,
+            'payment_mode_id': pay_mode_id if pay_mode_id else self.payment_mode_id.id
+        }
+        return inv_val_dict
```
ecino: All the else parts, if no value are provided, will simply write the same value as before. Isn't it better to not include it in the list of change in that case?
```suggestion
        inv_val_dict = {}
        if inv_lines:
            inv_val_dict["invoice_line_ids"] = inv_lines
        if due_date:
            inv_val_dict["date"] = due_date
        if ref:
            inv_val_dict["payment_reference"] = ref
        if pay_mode_id:
            inv_val_dict["payment_mode_id"] = pay_mode_id
        return {self.name: inv_val_dict}
```

recurring_contract/models/move.py
```diff
@@ -178,8 +177,101 @@ def find_sum(numbers, target, partial=None):
...
+            invoice.button_draft()
+            # Retrieve the value for a specific invoice
+            val_to_updt = updt_val.get(invoice.name)
+            # If the invoice is defined to_cancel we update it state
+            if "to_cancel" in val_to_updt["invoice_line_ids"]:
+                invoice._cancel_invoice()
+                continue
+            else:
+                invoice.write(val_to_updt)
+            invoice.action_post()
```
ecino: Based on my other change proposition, the dictionary could be empty and we could skip a not-needed write in some cases.
```suggestion
            # Retrieve the value for a specific invoice
            val_to_updt = updt_val.get(invoice.name)
            if val_to_updt:
                invoice.button_draft()
                # If the invoice is defined to_cancel we update it state
                if "to_cancel" in val_to_updt["invoice_line_ids"]:
                    invoice._cancel_invoice()
                    continue
                else:
                    invoice.write(val_to_updt)
                invoice.action_post()
```

recurring_contract/models/contract_group.py
```diff
@@ -80,17 +77,11 @@ def write(self, vals):
...
         for group in self:
-            # Get the method to apply changes
-            change_method = vals.get('change_method', group.change_method)
-            change_method = getattr(group, change_method)
-
             res = super(ContractGroup, group).write(vals) & res
-            change_method()
-
+            group._updt_invoices_cg(vals)
         return res
```
ecino: I don't think we are obliged to loop over the records. In my opinion working on the whole set directly will improve performance:
```
res = super().write(vals)
self._updt_invoices_cg(vals)
return res
```

recurring_contract/models/contract_group.py
```diff
@@ -304,3 +286,25 @@ def _setup_inv_data(self, journal, invoicer, contracts):
             'narration': "\n".join(contracts.mapped(lambda c: c.comment or ""))
         }
         return inv_data
+
+    def _updt_invoices_cg(self, vals):
+        """ method to update invoices on contrat group (cg)
+            :params vals dict of value that has been modified on the cg
+        """
+        self.ensure_one()
+        if "ref" in vals or "payment_mode_id" in vals or "advance_biling_months" in vals:
```
ecino: `advance_biling_months` is tested but not used at all in the update. I think it has no impact on the change of the invoices and could be removed from the test here, what do you think?

recurring_contract/models/contract_group.py
```diff
@@ -304,3 +286,25 @@ def _setup_inv_data(self, journal, invoicer, contracts):
             'narration': "\n".join(contracts.mapped(lambda c: c.comment or ""))
         }
         return inv_data
+
+    def _updt_invoices_cg(self, vals):
+        """ method to update invoices on contrat group (cg)
+            :params vals dict of value that has been modified on the cg
+        """
+        self.ensure_one()
```
ecino: We can handle the recordset all at once.
```suggestion
```

recurring_contract/models/contract_group.py
```diff
@@ -304,3 +286,25 @@ def _setup_inv_data(self, journal, invoicer, contracts):
...
+        if "ref" in vals or "payment_mode_id" in vals or "advance_biling_months" in vals:
+            invoices = self.env['account.move'].search([("partner_id", "=", self.partner_id.id),
```
ecino: In order to handle the whole recordset.
```suggestion
            invoices = self.env['account.move'].search([("partner_id", "in", self.mapped("partner_id").ids),
```

recurring_contract/models/move.py
```diff
@@ -178,8 +177,101 @@ def find_sum(numbers, target, partial=None):
...
+                                    'account_id': product.with_company(
+                                        contract_line.contract_id.company_id.id).property_account_income_id.id or False
+                                }
+                                ))
+                else:
+                    for line_id in line_ids:
+                        res.append((1, line_id, {"price_unit": amt}))
+
+        # Delete line if the product shouldn't be invoiced anymore
+        if len(inv_lines) < len(self.invoice_line_ids):
```
ecino: Maybe the user changed a product in the contract lines but didn't remove a line. In that case this test will fail to remove the old line with the old product, and will create just a new line. You should rather compare the set of products from the invoice_lines with the set of products from the contract_lines to make sure you remove all of those that are no longer in the contract.

recurring_contract/models/move.py
```diff
@@ -178,8 +177,101 @@ def find_sum(numbers, target, partial=None):
...
+            for line_id_to_del in lines_to_delete:
+                res.append((2, line_id_to_del, 0))
+        return res
+
+    def _cancel_invoice(self):
+        """
+        It cancels the invoice
+        """
+        self.ensure_one()
+        self.button_cancel()
```
ecino: I don't think this method brings any value. This could be done outside of this function, no?

recurring_contract/models/recurring_contract.py
```diff
@@ -657,3 +641,33 @@ def _filter_clean_invoices(self, since_date, to_date):
...
             invl_search.append(('due_date', '<=', to_date))
         return invl_search
+
+    def _updt_invoices_rc(self, vals):
+        """
+        It updates the invoices of a contract when the contract is updated
+
+        :param vals: the values that are being updated on the contract
+        """
+        self.ensure_one()
```
ecino: I think we can also deal with the recordset directly.
```suggestion
```

recurring_contract/models/recurring_contract.py
```diff
@@ -657,3 +641,33 @@ def _filter_clean_invoices(self, since_date, to_date):
...
+                    for line in self.contract_line_ids:
+                        line_data[line.product_id] = {"amt": line.subtotal, "contract": self}
+                    data_invs.update(
+                        inv._build_invoice_data(
+                            inv_lines=line_data,
+                            ref=self.group_id.ref,
+                            pay_mode_id=self.group_id.payment_mode_id
+                        )
+                    )
+                invoices.update_invoices(data_invs)
```
ecino: The invoices are already connected with the contract. Why do you filter the "sponsorship" products? As we are in the recurring_contract module, we cannot do that, it must be generic.
```suggestion
        if "contract_line_ids" in vals or "group_id" in vals:
            data_invs = dict()
            for contract in self:
                line_data = dict()
                for line in self.contract_line_ids:
                    line_data[line.product_id] = {"amt": line.subtotal, "contract": self}
                for inv in contract.mapped("invoice_line_ids.move_id").filtered(
                        lambda m: m.payment_state == "not_paid"):
                    data_invs.update(
                        inv._build_invoice_data(
                            inv_lines=line_data,
                            ref=contract.group_id.ref,
                            pay_mode_id=contract.group_id.payment_mode_id
                        )
                    )
            if data_invs:
                self.mapped("invoice_line_ids.move_id").update_invoices(data_invs)
```

ecino (changes requested):

recurring_contract/models/contract_group.py
```diff
@@ -80,17 +77,10 @@ def write(self, vals):
...
         res = True
-        for group in self:
-            # Get the method to apply changes
-            change_method = vals.get('change_method', group.change_method)
-            change_method = getattr(group, change_method)
-
-            res = super(ContractGroup, group).write(vals) & res
-            change_method()
-
+        res = super(ContractGroup, self).write(vals) & res
```
ecino, line 82: `res` is always True, it can be removed from line 81 as well.
```suggestion
        res = super().write(vals)
```

recurring_contract/models/contract_group.py
```diff
@@ -304,3 +285,21 @@ def _setup_inv_data(self, journal, invoicer, contracts):
...
         }
         return inv_data
+
+    def _updt_invoices_cg(self, vals):
+        """ method to update invoices on contrat group (cg)
+            :params vals dict of value that has been modified on the cg
+        """
+        if "ref" in vals or "payment_mode_id" in vals:
+            invoices = self.mapped("contract_ids.invoice_line_ids.move_id").filtered(
+                lambda i: i.payment_state == "not_paid")
```
ecino, line 295:
```suggestion
                lambda i: i.payment_state == "not_paid" and i.state != "cancel")
```

recurring_contract/models/move.py
```diff
@@ -178,8 +178,102 @@ def find_sum(numbers, target, partial=None):
...
+                                {
+                                    "price_unit": contract_line.amount,
+                                    "product_id": contract_line.product_id.id,
+                                    'name': contract_line.product_id.name,
+                                    'quantity': contract_line.quantity,
+                                    'contract_id': contract_line.contract_id.id,
+                                    'account_id': contract_line.product_id.with_company(
+                                        contract_line.contract_id.company_id.id).property_account_income_id.id or False
+                                }
+                                ))
```
ecino, line 265: Instead of coding the dictionary here, we could maybe move the method `def get_inv_lines_data(self):` declared in `recurring.contract` into the `recurring.contract.line` object instead so that we could call the method here with all the missing lines at once.
```suggestion
        missing_lines = contract.contract_line_ids.filtered(
            lambda l: l.product_id not in line_ids.mapped("product_id"))
        res.extend([(0, 0, line_vals) for line_vals in missing_lines.get_inv_lines_data()])
```

recurring_contract/models/move.py
```diff
@@ -178,8 +178,102 @@ def find_sum(numbers, target, partial=None):
...
+        """
+        res = []
+        line_ids = self.mapped("invoice_line_ids")
+        # Line to delete in the invoice
+        # If the invoice has different product than the contract we should delete the line
+        diff_product_invoices = [p for p in line_ids.mapped("product_id").ids if
+                                 p not in contract.contract_line_ids.mapped("product_id").ids]
+        if diff_product_invoices:
+            for diff_product_inv in diff_product_invoices:
+                res.append((2, line_ids.filtered(lambda l: l.product_id.id == diff_product_inv).id, 0))
```
ecino, line 246:
```suggestion
        contract_products = contract.contract_line_ids.mapped("product_id")
        diff_product_lines = line_ids.filtered(lambda l: l.product_id not in contract_products)
        res.extend([(2, line.id) for line in diff_product_lines])
```

recurring_contract/models/move.py
```diff
@@ -178,8 +178,102 @@ def find_sum(numbers, target, partial=None):
...
+        for line_id in line_ids:
+            cl = contract.mapped("contract_line_ids").filtered(lambda l: l.product_id == line_id.product_id)
+            if cl:
+                res.append((1, line_id.id,
+                            {
+                                "price_unit": cl.amount,
+                                "quantity": cl.quantity
+                            }
+                            )
+                           )
```
ecino, line 278:
```suggestion
        to_update = contract.contract_line_ids - (diff_product_lines + missing_lines)
        res.extend([(1, line.id, {
            "price_unit": line.amount,
            "quantity": line.quantity
        } for line in to_update])
```

recurring_contract/models/recurring_contract.py
```diff
@@ -657,3 +641,27 @@ def _filter_clean_invoices(self, since_date, to_date):
...
+        It updates the invoices of a contract when the contract is updated
+
+        :param vals: the values that are being updated on the contract
+        """
+        if "contract_line_ids" in vals or "group_id" in vals:
+            data_invs = {}
+            for contract in self:
+                line_data = {}
+                for line in self.contract_line_ids:
+                    line_data[line.product_id] = {"amt": line.subtotal, "contract": self}
```
ecino, line 656: `line_data` is never used!
```suggestion
```

recurring_contract/models/recurring_contract.py
```diff
@@ -657,3 +641,27 @@ def _filter_clean_invoices(self, since_date, to_date):
...
+                for inv in contract.mapped("invoice_line_ids.move_id").filtered(
+                        lambda m: m.payment_state == "not_paid"):
```
ecino, line 658:
```suggestion
                        lambda m: m.payment_state == "not_paid" and m.state != "cancel"):
```

ecino (changes requested):

PR #160, opened 2022-12-09
Direct debit mandate invoices cp-289

simgonzalez:
…_move


recurring_contract/views/account_move_view.xml
```diff
@@ -0,0 +1,22 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<!--
+  © 2013-2016 Akretion (Alexis de Lattre <alexis.delattre@akretion.com>)
+  License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+-->
+<odoo>
+    <record id="view_move_form" model="ir.ui.view">
+        <field name="name">add.mandate.on.account.move.form</field>
+        <field name="model">account.move</field>
+        <field name="inherit_id" ref="account_payment_partner.view_move_form"/>
```
ecino:
```suggestion
        <field name="inherit_id" ref="account_banking_mandate.view_move_form"/>
```

ecino (changes requested):
I know this is supposed to become a code that we share with the nordics, but I don't like adding a module dependency where it's not absolutely needed, it strongly reduces the genericity of the module in my opinion.

Because of this, I would prefer that we move this change in another module, (or create a new one if really needed). For instance, I propose `compassion-nordic/compassion_nordic_accounting` at the moment.

simgonzalez:
Closing pull request as it modify nordic now.

PR #161, opened 2022-12-09
14.0 fix price list

NoeBerdoz:
Related issue: [CP-229](https://compassion-ch.atlassian.net/browse/CP-229)

- Add the field pricelist in the sponsorship
- The pricelist relates to a company
- The company Sverige has 2 pricelist (EUR)
- When we select a partner in the sponsorship creation, the company field has to preselect the partner's company

The product subtotal/amount is updated in a related pull request: 

⚠️ Depending on pull request ⚠️ -> [PR compassion-module 1686](https://github.com/CompassionCH/compassion-modules/pull/1686)

### DIFF
--- a/recurring_contract/models/recurring_contract.py
+++ b/recurring_contract/models/recurring_contract.py
@@ -88,6 +88,8 @@ class RecurringContract(models.Model):
     company_id = fields.Many2one(
         'res.company',
         'Company',
+        # Show selection of all companies except Norden (id = 1)
+        domain="[('id', '!=', 1)]",
         required=True,
         default=lambda self: self.env.user.company_id.id, readonly=False
     )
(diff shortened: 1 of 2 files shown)


recurring_contract/models/recurring_contract.py
```diff
@@ -88,9 +88,18 @@ class RecurringContract(models.Model):
     company_id = fields.Many2one(
         'res.company',
         'Company',
+        # Show selection of all companies except Norden (id = 1)
+        domain="[('id', '!=', 1)]",
```
ecino: It's really a bad idea to hardcode this here. When we will use it for Switzerland, if our company has id 1 we won't be able to set it correctly! Please move this code in a nordic-specific module, for instance `compassion_nordic_accounting`

ecino (changes requested):

recurring_contract/models/recurring_contract.py
NoeBerdoz: Thank you very much for your review, I moved this code statement to the compassion_nordic_accounting module 
[Related PR compassion-nordic 9](https://github.com/CompassionCH/compassion-nordic/pull/19)

ecino (approved):

PR #162, opened 2022-12-23
CP-313 Migrate module to liberate invoice from an bank payment

simgonzalez:


ecino (changes requested):
The code looks good but in my opinion we can remove completely the "invoice.free.wizard" class and it's corresponding view, security access. It can be simply replaced by a server action calling the free_invoices method `free_payment_lines` in account.move class. An Action menu can be added which will call this action.

account_payment_line_free/wizards/__init__.py
```diff
@@ -0,0 +1,3 @@
+# -*- coding: utf-8 -*-
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+from . import invoice_free_wizard
```
ecino, line 3: This should be removed.

ecino (changes requested):

PR #166, opened 2023-01-10
CP-329 track_visibility for reference on payment option

simgonzalez:


recurring_contract/models/contract_group.py
```diff
@@ -49,7 +49,7 @@ class ContractGroup(models.Model):
     partner_id = fields.Many2one(
         'res.partner', 'Partner', required=True,
         ondelete='cascade', tracking=True, readonly=False)
-    ref = fields.Char('Reference')
+    ref = fields.Char('Reference', track_visibility='onchange')
```
ecino: This is the v12 way.
```suggestion
    ref = fields.Char('Reference', tracking=True)
```

ecino (changes requested):

ecino (approved):

PR #168, opened 2023-01-17
14.0 refactor inv

simgonzalez:


recurring_contract/models/move.py
```diff
@@ -202,24 +202,20 @@ def update_invoices(self, updt_val):
...
+                val_to_updt = updt_val[invoice.name]
+                # In case we modify the amount we want to test if the amount is zero
+                if "invoice_line_ids" in val_to_updt:
+                    tot_amt = sum(inv_line[2]["price_unit"] for inv_line in val_to_updt["invoice_line_ids"])
+                    if tot_amt == 0:
+                        invoice.button_cancel()
+                        continue
+                invoice.button_draft()
+                invoice.update(val_to_updt)
+                invoice.action_post()
```
ecino: To check the invoice total, I would rather do it after the update, to make sure that whatever the process is doing, it leaves with something valid. (ex: we could remove lines instead of updating the amount, no?).
```suggestion
                invoice.button_draft()
                invoice.update(val_to_updt)
                if invoice.amount_total:
                    invoice.action_post()
                else:
                    invoice.button_cancel()
```

ecino (changes requested):

ecino (approved):

PR #173, opened 2023-01-31
CP-239 Invoice refactor fixes

ecino:
- Remove useless recurring unit to avoid complexity
- Take recurring unit in consideration when generating invoices
- Remove account.move.line state field, which is a duplicate of already existing field parent_state
- ADD field for filtering contracts that needs invoice generation to happen. This could be used in the CRON of generation for instance or for users to check if everything runs smoothly.
- Various bug fixes

### DIFF
--- a/recurring_contract/models/recurring_contract.py
+++ b/recurring_contract/models/recurring_contract.py
@@ -229,17 +229,23 @@ class RecurringContract(models.Model):
         query = """
             SELECT COUNT(*)
             FROM account_move_line
-            WHERE contract_id = %s
+            WHERE contract_id = %s AND product_id = ANY(%s)
             AND due_date BETWEEN NOW() AND NOW() + INTERVAL '%s MONTHS'
         """
         for contract in self:
             if contract.state in ("waiting", "active"):
                 group = contract.group_id
-                self.env.cr.execute(query, [contract.id, group.advance_billing_months])
+                self.env.cr.execute(query, [
+                    contract.id,
+                    contract.mapped("contract_line_ids.product_id").ids,
+                    group.advance_billing_months]
+                )
                 number_invoices = self.env.cr.fetchone()[0]
                 if group.recurring_unit == "month":
-                    contract.missing_invoices = number_invoices < (group.advance_billing_months // group.recurring_unit)
+                    contract.missing_invoices = number_invoices < (
+                            group.advance_billing_months // group.recurring_value)
                 else:
+                    # This is the yearly case, there should be at least one invoice per year.
                     contract.missing_invoices = number_invoices == 0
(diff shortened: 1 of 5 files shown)


recurring_contract/models/recurring_contract.py
```diff
@@ -220,6 +224,26 @@ def _compute_months_paid(self):
...
+            WHERE contract_id = %s
+            AND due_date BETWEEN NOW() AND NOW() + INTERVAL '%s MONTHS'
+        """
+        for contract in self:
+            if contract.state in ("waiting", "active"):
+                group = contract.group_id
+                self.env.cr.execute(query, [contract.id, group.advance_billing_months])
+                number_invoices = self.env.cr.fetchone()[0]
+                if group.recurring_unit == "month":
+                    contract.missing_invoices = number_invoices < (group.advance_billing_months // group.recurring_unit)
```
simgonzalez: If we have some gift the boolean could be a false true ?

recurring_contract/models/recurring_contract.py
```diff
@@ -245,7 +269,7 @@ def write(self, vals):
         return res
 
     def unlink(self):
-        if self.filtered('start_date'):
+        if self.filtered('start_date') and not self.env.context.get("force_delete"):
```
simgonzalez, line 278: I know you didn't change that but is the start_date the condition for a validated contract ? Isn't the state the field we should test on ?

recurring_contract/models/recurring_contract.py
```diff
@@ -492,17 +517,22 @@ def _generate_invoices(self):
...
+                ("contract_id", "=", contract.id),
+                ("due_date", ">=", datetime.today()),
+                ("product_id", "in", contract.contract_line_ids.mapped("product_id").ids),
+                ("parent_state", "!=", "cancel")
+            ])
             _logger.info(f"Generating invoice for : {contract} {contract.reference}")
-            for inv_no in range(1, contract.group_id.advance_billing_months + 1):
+            recurring_unit = contract.group_id.recurring_unit
+            month_interval = contract.group_id.recurring_value * (1 if recurring_unit == "month" else 12)
+            for inv_no in range(1, contract.group_id.advance_billing_months + 1, month_interval):
```
simgonzalez, line 542: You could add a comment for the clarity :D

simgonzalez (changes requested):

recurring_contract/models/recurring_contract.py
ecino: The state can be changed at any time, the start_date is set once when we first validate the contract.

simgonzalez (approved):

PR #176, opened 2023-02-09
Cp 364 group invoices

simgonzalez:


recurring_contract/migrations/14.0.1.0.3/pre-migration.py
```diff
@@ -1,6 +0,0 @@
-
```
ecino, line 1: Why did you remove this migration? I don't find anything replacing it.

recurring_contract/migrations/14.0.1.0.5/post-migration.py
```diff
@@ -0,0 +1,47 @@
...
+
+@openupgrade.migrate()
+def migrate(env, version):
+    # Retrieve balance product
+    balance_product = env["product.product"].search([("name", "=", "Balance")])
+    if len(balance_product.ids) != 1:
+        _logger.warning("balance product is not unique")
+        return 1
+    # retrieve all contract line
+    contract_lines = env["recurring.contract"].search([("state", "in", ["waiting", "active"])]).mapped(
```
ecino: What about "draft" ? I think they should be included.

recurring_contract/models/product_pricelist_item.py
```diff
@@ -0,0 +1,28 @@
...
+
+class ProductPricelistItem(models.Model):
+    _inherit = 'product.pricelist.item'
+
+    def update_cl_amount(self):
+        for item in self:
+            contract_lines = self.env['recurring.contract.line'].search([
+                ('contract_id.pricelist_id', '=', item.pricelist_id.id),
+                ('product_id', '=', item.product_id.id),
+                ('contract_id.state', '=', ['waiting', 'active'])
```
ecino: Include also draft sponsorships

recurring_contract/models/product_pricelist_item.py
```diff
@@ -0,0 +1,28 @@
...
+            ])
+            for cl in contract_lines:
+                price = cl.contract_id.pricelist_id.get_product_price(cl.product_id,
+                                                                      cl.quantity,
+                                                                      cl.contract_id.partner_id)
+                if cl.amount != price:
+                    cl.amount = price
+        return {
```
ecino, line 20: As the process will be run in the background, I don't think anybody will get this notification. Unless some people will run the method manually somewhere in the user interface.

recurring_contract/views/recurring_contract_view.xml
```diff
@@ -31,38 +32,36 @@
...
+                    <button name="contract_waiting" class="oe_highlight" string="Validate" states="draft"
+                            type="object"/>
                     <button name="action_cancel_draft" type="object" string="Set back to draft" states="cancelled"/>
                     <button name="%(action_end_contract)d" type="action" string="Terminate" class="oe_highlight"
                             attrs="{'invisible': ['|', ('state', 'in', ['terminated', 'cancelled']), ('end_date', '!=', False)]}"/>
                 </header>
                 <sheet>
                     <div class="oe_button_box" name="button_box">
-                        <button name="button_generate_invoices" type="object" class="oe_inline oe_stat_button"
-                                icon="fa-refresh" string="Generate invoices" states="waiting,active"/>
```
ecino: I'm not sure it's good to remove the button from the contract. It means that the user will always need to go in the group when he wants to generate invoices (one click more). This button could simply be a shortcut calling the same method. I think users would complain if the button is removed.

ecino (changes requested):

recurring_contract/migrations/14.0.1.0.3/pre-migration.py
simgonzalez: I moved it in compassion-module :) it was a migration on a field defined in the sponsorship not in the recurring_contract

recurring_contract/migrations/14.0.1.0.5/post-migration.py
simgonzalez: The draft state is currently not used, when we call the generation of invoices the contract is in waiting state,

recurring_contract/models/product_pricelist_item.py
simgonzalez: The draft state is currently not used, when we call the generation of invoices the contract is in waiting state,

recurring_contract/models/product_pricelist_item.py
simgonzalez: It's for the future :) In case we want this update to be called manually we have a feedback of what happened :D

recurring_contract/views/recurring_contract_view.xml
simgonzalez: I agree with you but putting that button on the contract is quite confusing because from the contract you'll generate invoices for the group. So for me we could put a button on the partner in addition of the contract group but we shouldn't put it on the contract.

recurring_contract/models/product_pricelist_item.py
ecino: Yes but if the price is not updated and then a user validates the sponsorship without checking, the sponsorship will not have the same amount as it should when invoice generation will start.

recurring_contract/views/recurring_contract_view.xml
ecino: It has been like that forever and no one was ever confused. The partner view already has hundred of buttons, I think it's already a confused view where I don't wan't to include every possible action. In general, when working with sponsorships, we are in the contract view and it's from there that we need the invoices. Contract group is a rarely accessed view.

recurring_contract/migrations/14.0.1.0.5/post-migration.py
ecino: I think here also including draft is needed.

recurring_contract/models/contract_group.py
```diff
@@ -240,7 +240,8 @@ def _generate_invoices(self):
             # Compute the interval of months there should be between each invoice (set in the contract group)
             recurring_unit = pay_opt.recurring_unit
             month_interval = pay_opt.recurring_value * (1 if recurring_unit == "month" else 12)
-            for inv_no in range(1, pay_opt.advance_billing_months + 1, month_interval):
+            curr_month = self.env["ir.config_parameter"].sudo().get_param(f"compassion_nordic_accounting.do_generate_curr_month_{self.env.company.id}", 0)
+            for inv_no in range(0 if curr_month else 1, pay_opt.advance_billing_months + 1, month_interval):
```
ecino: `def get_param` is declared in module `message_center_compassion` and is thus not accessible from within this module that has no dependency on it.

Can the company be read from the environment or should it be taken from the contract_group or contract object?

```suggestion
            param_string = f"compassion_nordic_accounting.do_generate_curr_month_{self.env.company.id}"
            curr_month = self.env["ir.config_parameter"].sudo().default_get([param_string]).get(param_string, False)
            for inv_no in range(0 if curr_month else 1, pay_opt.advance_billing_months + 1, month_interval):
```

recurring_contract/models/recurring_contract.py
```diff
@@ -233,8 +231,14 @@ def create(self, vals):
...
-
-        return super().create(vals)
+        res = super().create(vals)
+        # Define if the invoices should temporarily be suspended
+        today = fields.Date.today()
+        day_block_inv = self.env["ir.config_parameter"].sudo().get_param(
+            f"compassion_nordic_accounting.inv_block_day_{self.env.company.id}", 31)
+        if today.day > day_block_inv:
+            self.mapped("group_id").write({"invoice_suspended_until": today + relativedelta(months=1)})
+        return res
```
ecino: I would rather implement this logic in a `@api.onchange` method or as default value method, so that the date can be visible in the user interface before saving the document. Otherwise the date will be set after we save the contract and the user may not notice (He may still want to be in control of that).

recurring_contract/models/res_config_settings.py
```diff
@@ -0,0 +1,55 @@
...
+        res["do_generate_curr_month"] = self.get_param_multi_company("compassion_nordic_accounting.do_generate_curr_month")
+        return res
+
+    def set_values(self):
+        self._set_param_multi_company("compassion_nordic_accounting.invoice_block_day", self.inv_block_day)
+        self._set_param_multi_company("compassion_nordic_accounting.do_generate_curr_month", str(self.do_generate_curr_month))
+        super().set_values()
+
+    def get_param_multi_company(self, par_name):
+        return self.env["ir.config_parameter"].sudo().get_param(f"{par_name}_{self.env.company.id}", 0)
```
ecino: The `get_param` method is not accessible from this module (see previous comment). This should be removed.

ecino (changes requested):

ecino (approved):

PR #180, opened 2023-02-23
CP-384 FIX invoice cancellation at contract termination

ecino:
- Redo paid invoices cancellation, the process was the old version from 12.0 which was no longer adapted. It simplifies the process by not splitting payments but just make use of the reconcile function which correctly allocates partial amounts in case the payment is greater than the invoice.
- Make use of the invoice blocking day to decide whether cancelling the current month invoice or not.

### DIFF
--- a/recurring_contract/models/move.py
+++ b/recurring_contract/models/move.py
@@ -92,7 +92,7 @@ class AccountMove(models.Model):
         mvl_obj = self.env['account.move.line']
         for partner in self.mapped('partner_id'):
             invoices = self.filtered(lambda i: i.partner_id == partner)
-            past_invoices = invoices.filtered(lambda i: i.date <= today)
+            past_invoices = invoices.filtered(lambda i: i.invoice_date_due <= today)
             past_lines = past_invoices.mapped('line_ids').filtered("debit")
             past_amount = sum(past_invoices.mapped('amount_total'))
             future_invoices = invoices - past_invoices
--- a/recurring_contract/models/move.py
+++ b/recurring_contract/models/move.py
@@ -84,123 +84,48 @@ class AccountMove(models.Model):
     def reconcile_after_clean(self):
         """
         Called after clean invoices. If invoices can be reconciled
-        with open payment, this will split the payment into three amounts :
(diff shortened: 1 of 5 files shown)


recurring_contract/models/move.py
```diff
@@ -84,124 +84,49 @@ def action_invoice_re_open(self):
...
-        cancel_invoices = self.filtered(lambda i: i.state == 'cancel')
-        cancel_invoices.button_draft()
-        cancel_invoices.action_post()
         today = date.today()
+        mvl_obj = self.env['account.move.line']
         for partner in self.mapped('partner_id'):
             invoices = self.filtered(lambda i: i.partner_id == partner)
-            past_invoices = invoices.filtered(lambda i: i.invoice_date <= today)
-            future_invoices = invoices - past_invoices
+            past_invoices = invoices.filtered(lambda i: i.date <= today)
```
simgonzalez:
```suggestion
            past_invoices = invoices.filtered(lambda i: i.invoice_date_due <= today)
```
The due_date  defines when the invoice shoud be paid for so i think that's a better check to know if an invoice is in the past. Moreover, it's more explicit than date field.

recurring_contract/models/recurring_contract.py
```diff
@@ -440,48 +444,68 @@ def _cancel_invoices(self):
...
+
+        # Unreconcile paid invoices
+        move_lines |= reconciles.mapped('reconciled_line_ids')
+        move_lines.remove_move_reconcile()
+        paid_invoices = inv_lines_paid.mapped("move_id")
+        paid_invoices.button_draft()
+        for inv in paid_invoices:
+            inv.write({"invoice_line_ids": [(2, invl.id) for invl in inv_lines_paid.exists() if invl.move_id == inv]})
+        paid_invoices.action_post()
+        paid_invoices.with_company(self.mapped("company_id").id).reconcile_after_clean()
```
simgonzalez:
```suggestion
        paid_invoices.with_company(self[0].company_id.id).reconcile_after_clean()
```

simgonzalez (changes requested):

PR #184, opened 2023-02-28
FIX for contract invoices update

ecino:
- Update current month invoices so that if we make a change in contract during the month, the invoice for the current month is updated as well.
- Make an exception if we cross the invoice blocking date. In that case, update only starting from the next month.
- Refactor method name


simgonzalez (approved):
Invoice block day was a parameter for the creation of new contract (that's what the helper indicated). Maybe updating the helper would make it less confusing. Except that, the changes are really good.

ecino:
Thanks. I updated the help message.

PR #187, opened 2023-03-07
CP-405 Further refactor for invoicing system

ecino:
- [FIX] More granularity when updating a sponsorship. Prevents a change in one sponsorship to affect all other invoice lines in the same invoice.
- Move code to better isolate gift functionality inside sponsorship_compassion module instead of recurring_contract.


recurring_contract/models/contract_group.py
```diff
@@ -384,17 +384,9 @@ def _updt_invoices_cg(self, vals):
...
-                data_invs = dict()
-                for inv in invoices:
-                    data_invs.update(
-                        inv._build_invoice_data(
-                            ref=vals.get("ref"),
-                            pay_mode_id=vals.get("payment_mode_id")
-                        )
-                    )
+            invoices = self.mapped("active_contract_ids.open_invoice_ids")
+            data_invs = invoices._build_invoice_data(ref=vals.get("ref"), pay_mode_id=vals.get("payment_mode_id"))
```
simgonzalez:
```suggestion
            data_invs = invoices._build_invoices_data(ref=vals.get("ref"), pay_mode_id=vals.get("payment_mode_id"))
```
Name should be adapted to new behaviour

recurring_contract/models/move.py
```diff
@@ -151,83 +149,65 @@ def update_open_invoices(self, updt_val):
                 else:
                     invoice.button_cancel()
 
-    def _build_invoice_data(self, contract=False, invoice_date=False, ref=False, pay_mode_id=False,
+    def _build_invoice_data(self, contracts=False, invoice_date=False, ref=False, pay_mode_id=False,
```
simgonzalez:
```suggestion
    def _build_invoices_data(self, contracts=False, invoice_date=False, ref=False, pay_mode_id=False,
```
Name should be adapted to new behaviour

recurring_contract/models/move_line.py
```diff
@@ -55,3 +55,28 @@ def group_reconcile(self, matched_lines, credit_or_debit="debit"):
...
             return False
         return (self | selected_lines).reconcile()
+
+    def _update_invoice_lines_from_contract(self, modified_contract):
+        """
+        Takes the contract as the source to generate a write command for updating the invoice line
+        :param modified_contract: <recurring.contract> record
+        :return: list of tuples for ORM write
+        """
+        res = []
```
simgonzalez, line 66:
```suggestion
        modified_contract.ensure_one()
        res = []
```

recurring_contract/models/recurring_contract.py
```diff
@@ -514,17 +521,18 @@ def _updt_invoices_rc(self, vals):
...
-                            ref=contract.group_id.ref,
-                            pay_mode_id=contract.group_id.payment_mode_id
-                        )
-                    )
-            if data_invs:
-                self.mapped("invoice_line_ids.move_id").update_open_invoices(data_invs)
+        data_invs = {}
+        if "contract_line_ids" in vals:
+            data_invs = self.mapped("open_invoice_ids")._build_invoice_data(contracts=self)
+        if "group_id" in vals:
```
simgonzalez, line 527: If we update the partner_id will it update group_id ? Is this why we don't catch partner_id there ?

recurring_contract/models/recurring_contract.py
```diff
@@ -514,17 +521,18 @@ def _updt_invoices_rc(self, vals):
...
+            group = self.mapped("group_id")
+            group.ensure_one()
+            group_data = self.mapped("open_invoice_ids")._build_invoice_data(
+                ref=group.ref, pay_mode_id=group.payment_mode_id)
+            for inv_name, inv_data in group_data.items():
+                if inv_name in data_invs:
+                    data_invs[inv_name].update(inv_data)
+                else:
+                    data_invs[inv_name] = inv_data
```
simgonzalez, line 536:
```suggestion
                data_invs[inv_name] = inv_data
```
We could simplify this code by writing multiple times the same invoices in the dictionnary for updates. Although that would call multiple times postgresql for the same object.

recurring_contract/models/recurring_contract.py
```diff
@@ -514,17 +521,18 @@ def _updt_invoices_rc(self, vals):
...
-                    )
-            if data_invs:
-                self.mapped("invoice_line_ids.move_id").update_open_invoices(data_invs)
+        data_invs = {}
+        if "contract_line_ids" in vals:
+            data_invs = self.mapped("open_invoice_ids")._build_invoice_data(contracts=self)
+        if "group_id" in vals:
+            group = self.mapped("group_id")
+            group.ensure_one()
+            group_data = self.mapped("open_invoice_ids")._build_invoice_data(
```
simgonzalez:
```suggestion
            group_data = self.mapped("open_invoice_ids")._build_invoices_data(
```

recurring_contract/models/recurring_contract.py
```diff
@@ -514,17 +521,18 @@ def _updt_invoices_rc(self, vals):
...
-                            contract=contract,
-                            ref=contract.group_id.ref,
-                            pay_mode_id=contract.group_id.payment_mode_id
-                        )
-                    )
-            if data_invs:
-                self.mapped("invoice_line_ids.move_id").update_open_invoices(data_invs)
+        data_invs = {}
+        if "contract_line_ids" in vals:
+            data_invs = self.mapped("open_invoice_ids")._build_invoice_data(contracts=self)
```
simgonzalez:
```suggestion
            data_invs = self.mapped("open_invoice_ids")._build_invoices_data(contracts=self)
```

recurring_contract/tests/test_recurring_contract.py
```diff
@@ -95,7 +95,7 @@ def test_build_invoice_data_contract(self):
         Asserts that the returned dictionary contains the correct invoice data and that the invoice line ids are in the
         right format
         """
-        result = self.invoices[0]._build_invoice_data(contract=self.contract)
+        result = self.invoices[0]._build_invoice_data(contracts=self.contract)
```
simgonzalez:
```suggestion
        result = self.invoices[0]._build_invoices_data(contracts=self.contract)
```

simgonzalez (changes requested):

recurring_contract/models/recurring_contract.py
ecino: Yes, when changing the partner, the group also changes because we cannot use a group from another partner. Note that in the previous code version, the `partner_id` field was also not in the list of trigger fields here.

recurring_contract/models/recurring_contract.py
ecino: No, a dictionary cannot contain multiple times the same key. Writing inside without checking would override the previous values.

simgonzalez (approved):

PR #188, opened 2023-04-04
T0103 invoice payment order restrictions.

simgonzalez:
Add restrictions for the payment on payment order.

### DIFF
--- a/invoice_restrictions/__manifest__.py
+++ b/invoice_restrictions/__manifest__.py
@@ -8,8 +8,8 @@
 #                        /_/
 #                            in Jesus' name
 #
-#    Copyright (C) 2014-2019 Compassion CH (http://www.compassion.ch)
-#    @author: Cyril Sester <csester@compassion.ch>, Emanuel Cino
+#    Copyright (C) 2014-2023 Compassion CH (http://www.compassion.ch)
+#    @author: Emanuel Cino
 #
 #    This program is free software: you can redistribute it and/or modify
 #    it under the terms of the GNU Affero General Public License as
@@ -25,9 +25,8 @@
 #    along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #
 ##############################################################################
-# pylint: disable=C8101
 {
-    'name': 'Recurring contract',
+    'name': 'Restrictions on invoices',
     'summary': 'Contract for recurring invoicing',
     'version': "14.0.1.0.5",
     'license': 'AGPL-3',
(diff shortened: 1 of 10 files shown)


recurring_contract/models/move.py
```diff
@@ -139,6 +139,9 @@ def update_open_invoices(self, updt_val):
         date_selection = date_selection.replace(day=1)
         for invoice in self.filtered(lambda i: i.state != "cancel" and i.payment_state != "paid"
                                      and i.invoice_date_due >= date_selection):
+            # When the invoice has a payment order
+            if invoice.payment_order_id.filtered(lambda p: p.state not in ["draft", "open"]):
+                continue
```
ecino: This could simply be included in the filtered conditions in the loop above.

invoice_restrictions/models/account_move.py
```diff
@@ -0,0 +1,22 @@
...
+
+from odoo import models
+
+
+class AccountMove(models.Model):
+    _name = 'account.move'
+    _inherit = 'account.move'
+
+    def write(self, vals):
+        if self.mapped('payment_order_id').filtered(lambda p: p.state not in ["draft", "open"]):
```
ecino: Should we filter for specific vals change only? Wouldn't some changes without impact on the payment order still be possible? I feel this global way is too preventive and might raise errors for cases that are not problematic in the end.

invoice_restrictions/__manifest__.py
```diff
@@ -0,0 +1,42 @@
...
+#       ______ Releasing children from poverty      _
+#      / ____/___  ____ ___  ____  ____ ___________(_)___  ____
+#     / /   / __ \/ __ `__ \/ __ \/ __ `/ ___/ ___/ / __ \/ __ \
+#    / /___/ /_/ / / / / / / /_/ / /_/ (__  |__  ) / /_/ / / / /
+#    \____/\____/_/ /_/ /_/ .___/\__,_/____/____/_/\____/_/ /_/
+#                        /_/
+#                            in Jesus' name
+#
+#    Copyright (C) 2014-2019 Compassion CH (http://www.compassion.ch)
+#    @author: Cyril Sester <csester@compassion.ch>, Emanuel Cino
```
ecino: It's not really reflecting the reality.

ecino (changes requested):

simgonzalez:
The rules must be rethought because it's not as easy as it looks and we have to define rules for exceptional case. Like taking out of the payment order the move and changing his payment_state from not paid to paid. and ...

invoice_restrictions/models/account_move.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+class AccountMove(models.Model):
+    _name = 'account.move'
+    _inherit = 'account.move'
+
+    def write(self, vals):
+        move_to_modify = self.filtered('line_ids.payment_line_ids')
+        if move_to_modify:
+            payment_line_ids = move_to_modify.mapped('line_ids.payment_line_ids')
```
simgonzalez: test the status in vals.

ecino (approved):

PR #189, opened 2023-04-20
T103 fix: invoices aren't modified.

simgonzalez:
Payment_order_id might be null and make the move not to be modified when it should.

### DIFF
--- a/invoice_restrictions/models/account_move.py
+++ b/invoice_restrictions/models/account_move.py
@@ -15,26 +15,20 @@ class AccountMove(models.Model):
     _inherit = 'account.move'
 
     def write(self, vals):
+        is_write_paystate = False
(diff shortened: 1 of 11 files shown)


ecino (approved):

invoice_restrictions/models/account_move.py
```diff
@@ -15,26 +15,21 @@ class AccountMove(models.Model):
     _inherit = 'account.move'
 
     def write(self, vals):
+        is_write_paystate = False
+        if vals.get("payment_state"):
+          is_write_paystate = True
```
ecino:
```suggestion
        is_write_paystate = vals.get("payment_state")
```

ecino (changes requested):

ecino (approved):

PR #193, opened 2023-05-04
Prob on the param do curr month

simgonzalez:


NoeBerdoz (approved):

NoeBerdoz:
Why was there two commits for the one change to make? @simgonzalez 
Would be cleaner to have only one commit to merge, [878c42 commit](https://github.com/CompassionCH/compassion-accounting/pull/193/commits/878c42134672c403889cb8c4eebf31ff7d3060fe) is not modifiying any codebase

PR #208, opened 2023-06-12
T0186 move fiscal year not working as expected

simgonzalez:
We should adapt it to modify the move in draft.


account_move_periodic_accounting_transfer/models/res_company.py
```diff
@@ -26,15 +26,20 @@ def _validate_fiscalyear_lock(self, values):
...
-                    analytic_lines.write({'date': first_day_in_next_fy})
+                    moves = open_invoices.sudo()
+                    for move in moves:
+                        move.button_draft()
+                        move.write({
+                            'date': first_day_in_next_fy
+                        })
+                        move.action_post()
+                        analytic_lines = move.mapped('line_ids.analytic_line_ids')
+                        analytic_lines.write({'date': first_day_in_next_fy})
```
ecino: This may not longer be needed, because the draft and post actions should recreate the analytic lines with the updated date (in my opinion). Can you test and remove it if it's working?

Otherwise the change may slow down a lot the process. Are we obliged to cancel and re-post the move in order to change the date?

ecino (approved):

PR #211, opened 2023-06-30
T0351 adapt completion rules

simgonzalez:
Adapt completion rules to be editable in user interface


account_statement_completion/models/completion_rules.py
```diff
@@ -28,6 +29,29 @@ class StatementCompletionRule(models.Model):
...
+                    #-------------------------------
+                    # stmts_vals: Values of the statements as a list of dict
+                    # stmt_line: Values of the statement line as a dict
+                    # line_amount: stmt_line amount
+                    # ref: stmt_line reference
+                    # AccMove: Odoo model "account.move"
+                    # AccMoveLine: Odoo model "account.move.line"
+                    # Partner: Odoo model "res.partner".
+                    # Payment: Odoo model "account.payment"
+                    # PaymentLine: Odoo model "account.payment.line"
```
ecino: Why not simply providing the `env` variable in the eval context? That way the functions can use any model needed with `env["account.move"]` for instance.

account_statement_completion/models/completion_rules.py
```diff
@@ -62,71 +85,29 @@ def auto_complete(self, stmts_vals, stmt_line):
...
-            method = getattr(self, rule.function_to_call)
-            result = method(stmts_vals, stmt_line)
+            eval_dict = rule._get_base_dict(stmts_vals, stmt_line)
+            safe_eval.safe_eval(
+                rule.python_completion_rule,
+                eval_dict,
+                mode="exec",
+                nocopy=True
+            )
+            result = eval_dict.get("result", {})
```
ecino: When checking and discussing with David, the way of processing is not good (it wasn't good even before your changes). Instead of looping through the rules and applying only the first that has some results, we should apply every rule on every bank statement line.

Moreover, if `safe_eval` is not doing a copy of the `stmts_vals`and `stmt_line` variables, we don't need to return any result and can adapt the [calling code here](https://github.com/compassionch/compassion-accounting/blob/d4f7303244f31275b5de6cd184fd4aea783b4730/account_statement_completion/models/bank_statement_import.py#L28). The safe_eval method can only insert its results inside the input variables.

```
    @api.model
    def _complete_stmts_vals(self, stmts_vals, journal, account_number):
        """
        Use completion rules in journal to find missing partners.
        """
        stmts_vals = super()._complete_stmts_vals(
            stmts_vals, journal, account_number)
        if journal.completion_rules:
            for st_vals in stmts_vals:
                journal.completion_rules.auto_complete(st_vals)
```

Then here in the method it would be something like this:

```
    def auto_complete(self, stmt_vals):
        """This method will execute all related rules, in their sequence order,
        to retrieve all the values returned by the first rules that will match.
        :param stmt_vals: dict with bank statement values
        return:
            A dict of values for the bank statement line or {}
           {'partner_id': value,
            'account_id': value,
            ...}
        """
        stmt_lines_vals = stmt_vals.get("transactions", list())
        for stmt_line_vals in stmt_lines_vals:
            for rule in self.sorted(key=lambda r: r.sequence):
                eval_dict = rule._get_base_dict(stmt_vals, stmt_line_vals)
                safe_eval.safe_eval(
                    rule.python_completion_rule,
                    eval_dict,
                    mode="exec",
                    nocopy=True
                )
                if eval_dict.get("result"):
                    break
```

account_statement_completion/models/completion_rules.py
```diff
@@ -28,6 +29,29 @@ class StatementCompletionRule(models.Model):
...
+                    # ref: stmt_line reference
+                    # AccMove: Odoo model "account.move"
+                    # AccMoveLine: Odoo model "account.move.line"
+                    # Partner: Odoo model "res.partner".
+                    # Payment: Odoo model "account.payment"
+                    # PaymentLine: Odoo model "account.payment.line"
+
+                    # Available compute variables:
+                    #-------------------------------
+                    # result: returned value have to be set in the variable 'result'
```
ecino:
```suggestion
                    # result: True if the reconcilation rule found any relevant data
```

account_statement_completion/migrations/14.0.1.0.1/pre-migration.py
```diff
@@ -0,0 +1,13 @@
+from openupgradelib import openupgrade
```
ecino: I doubt this migration is useful. All existing rules will be invalid if they contain only one string. If we remove the migration it will simply create a new column in the database for the rule code, and remove the previous field which is useless now.

ecino (changes requested):

account_statement_completion/models/completion_rules.py
```diff
@@ -37,96 +61,55 @@ class StatementCompletionRule(models.Model):
...
-            ('name', '=', ref)
-        ], limit=1, order='date desc')
-        if payment_line:
-            res['partner_id'] = payment_line.partner_id.\
-                commercial_partner_id.id
-        return res
+    def _get_base_dict(self, stmts_vals, stmt_line):
+        """ Return the values usable in the code template """
+        return {
+            "result": {},
```
ecino:
```suggestion
            "result": False,
```

account_statement_completion/models/completion_rules.py
```diff
@@ -28,6 +29,29 @@ class StatementCompletionRule(models.Model):
...
+                    # stmts_vals: Values of the statements as a list of dict
+                    # stmt_line: Values of the statement line as a dict
+                    # line_amount: stmt_line amount
+                    # ref: stmt_line reference
+                    # env: environment
+                    # AccMove: Odoo model "account.move"
+                    # AccMoveLine: Odoo model "account.move.line"
+                    # Partner: Odoo model "res.partner".
+                    # Payment: Odoo model "account.payment"
+                    # PaymentLine: Odoo model "account.payment.line"
```
ecino: I would remove all these variables from the evaluation context. They can be inferred when needed in the code using the `env` variable.

account_statement_completion/models/completion_rules.py
```diff
@@ -28,6 +29,29 @@ class StatementCompletionRule(models.Model):
...
+
+                    # Available compute variables:
+                    #-------------------------------
+                    # result: True if the reconcilation rule found any relevant data
```
ecino:
```suggestion
                    # result: True if the reconcilation rule found any relevant data.
                    #             This will prevent any subsequent rule to be executed on the same line.
```

account_statement_completion/data/data.xml
```diff
@@ -1,22 +1,66 @@
...
+        # We check only for debit entries
+        if amount &lt; 0:
+            invoices = AccMove.search(
+                [('move_type', '=', 'in_invoice'), ('state', '=', 'posted'),
+                 ('amount_total', '=', abs(amount))])
+            result = {}
+            if invoices:
+                if len(invoices) == 1:
+                    partner = invoices.partner_id
+                    stmt_line['partner_id'] = partner.commercial_partner_id.id
```
ecino, line 19:
```suggestion
                    stmt_line['partner_id'] = partner.commercial_partner_id.id
                    result = True
```

account_statement_completion/data/data.xml
```diff
@@ -1,22 +1,66 @@
...
+                else:
+                    partner = invoices[0].partner_id
+                    for invoice in invoices:
+                        if invoice.partner_id.id != partner.id:
+                            logger.warning(
+                                f"Line named {st_line['name']} (Ref:{st_line['ref']})"
+                                f" was matched by "
+                                'more than one invoice while looking on open'
+                                ' supplier invoices')
+                    stmt_line['partner_id'] = partner.commercial_partner_id.id
```
ecino, line 30:
```suggestion
                    stmt_line['partner_id'] = partner.commercial_partner_id.id
                    result = True
```

account_statement_completion/data/data.xml
```diff
@@ -1,22 +1,66 @@
...
+            partner = None
+
+            # Search move lines
+            move_lines = AccMoveLine.search(
+                [('ref', '=', ref), ('partner_id', '!=', None)])
+            if move_lines:
+                partner = move_lines[0].partner_id
+
+            if partner:
+                stmt_line['partner_id'] = partner.commercial_partner_id.id
```
ecino, line 51:
```suggestion
                stmt_line['partner_id'] = partner.commercial_partner_id.id
                result = True
```

account_statement_completion/data/data.xml
```diff
@@ -1,22 +1,66 @@
...
-        <field name="function_to_call">get_from_payment_line</field>
+        <field name="python_completion_rule">
+        """ Search in account.payment.line """
+        if ref:
+            payment_line = PaymentLine.search([
+                ('name', '=', ref)
+            ], limit=1, order='date desc')
+            if payment_line:
+                stmt_line['partner_id'] = payment_line.partner_id.\
+                    commercial_partner_id.id
```
ecino:
```suggestion
                stmt_line['partner_id'] = payment_line.partner_id.commercial_partner_id.id
                result = True
```

account_statement_completion/models/completion_rules.py
```diff
@@ -37,96 +61,55 @@ class StatementCompletionRule(models.Model):
...
+            "stmts_vals": stmts_vals,
+            "stmt_line": stmt_line,
+            "line_amount": int(stmt_line["amount"]),
+            "ref": stmt_line.get('ref'),
+            "env": self.env,
+            "AccMove": self.env["account.move"],
+            "AccMoveLine": self.env["account.move.line"],
+            "Partner": self.env["res.partner"],
+            "Payment": self.env["account.payment"],
+            "PaymentLine": self.env["account.payment.line"]
```
ecino:
```suggestion
```

account_statement_completion/models/completion_rules.py
```diff
@@ -37,96 +61,55 @@ class StatementCompletionRule(models.Model):
...
-                commercial_partner_id.id
-        return res
+    def _get_base_dict(self, stmts_vals, stmt_line):
+        """ Return the values usable in the code template """
+        return {
+            "result": {},
+            "stmts_vals": stmts_vals,
+            "stmt_line": stmt_line,
+            "line_amount": int(stmt_line["amount"]),
+            "ref": stmt_line.get('ref'),
```
ecino:
```suggestion
```

account_statement_completion/data/data.xml
```diff
@@ -1,22 +1,66 @@
...
+<odoo noupdate="1">
     <record id="bank_statement_completion_rule_supplier_amount" model="account.statement.completion.rule">
         <field name="name">Match from line amount (based on supplier invoice amount)</field>
         <field name="sequence">60</field>
-        <field name="function_to_call">get_from_amount</field>
+        <field name="python_completion_rule">
+        """ If line amount match an open supplier invoice,
+        update partner and account. """
+        # We check only for debit entries
+        if amount &lt; 0:
```
ecino:
```suggestion
        # We check only for debit entries
        amount = stmt_line.get("amount", 0)
        if amount &lt; 0:
```

account_statement_completion/data/data.xml
```diff
@@ -1,22 +1,66 @@
...
+        </field>
     </record>
 
     <record id="bank_statement_completion_rule_move_line_ref" model="account.statement.completion.rule">
         <field name="name">From line reference (based on previous move_line references)</field>
         <field name="sequence">70</field>
-        <field name="function_to_call">get_from_move_line_ref</field>
+        <field name="python_completion_rule">
+        ''' Update partner if same reference is found '''
+        if ref:
```
ecino:
```suggestion
        ''' Update partner if same reference is found '''
        ref = stmt_line.get("ref")
        if ref:
```

account_statement_completion/data/data.xml
```diff
@@ -1,22 +1,66 @@
...
+        </field>
     </record>
 
     <record id="bank_statement_completion_rule_payment_line" model="account.statement.completion.rule">
         <field name="name">From line reference (based on payment line reference)</field>
         <field name="sequence">10</field>
-        <field name="function_to_call">get_from_payment_line</field>
+        <field name="python_completion_rule">
+        """ Search in account.payment.line """
+        if ref:
```
ecino:
```suggestion
        """ Search in account.payment.line """
        ref = stmt_line.get("ref")
        if ref:
```

ecino (changes requested):

account_statement_completion/models/completion_rules.py
simgonzalez: Does it change a lot performance wise ? If not giving an easy accessible base for the "users" seems quite convenient to me.

account_statement_completion/models/completion_rules.py
simgonzalez: Same as the env variable, to me giving this access isn't a bad idea.

account_statement_completion/models/completion_rules.py
ecino: The users are still technical skilled people that should know about the "env" in Odoo. Using that is easier than having to know every existing variable in the context. Please make the change.

account_statement_completion/data/data.xml
```diff
@@ -1,22 +1,72 @@
...
         <field name="name">Match from line amount (based on supplier invoice amount)</field>
         <field name="sequence">60</field>
-        <field name="function_to_call">get_from_amount</field>
+        <field name="python_completion_rule">
+        """ If line amount match an open supplier invoice,
+        update partner and account. """
+        # We check only for debit entries
+        amount = int(stmt_line["amount"])
+        if amount &lt; 0:
+            invoices = self.env["account.move"].search(
```
ecino:
```suggestion
            invoices = env["account.move"].search(
```

account_statement_completion/data/data.xml
```diff
@@ -1,22 +1,72 @@
...
         <field name="sequence">70</field>
-        <field name="function_to_call">get_from_move_line_ref</field>
+        <field name="python_completion_rule">
+        ''' Update partner if same reference is found '''
+        ref = stmt_line.get('ref')
+        if ref:
+            partner = None
+
+            # Search move lines
+            move_lines = self.env["account.move.line"].search(
```
ecino:
```suggestion
            move_lines = env["account.move.line"].search(
```

account_statement_completion/data/data.xml
```diff
@@ -1,22 +1,72 @@
...
 
     <record id="bank_statement_completion_rule_payment_line" model="account.statement.completion.rule">
         <field name="name">From line reference (based on payment line reference)</field>
         <field name="sequence">10</field>
-        <field name="function_to_call">get_from_payment_line</field>
+        <field name="python_completion_rule">
+        """ Search in account.payment.line """
+        ref = stmt_line.get('ref')
+        if ref:
+            payment_line = self.env["account.payment.line"].search([
```
ecino:
```suggestion
            payment_line = env["account.payment.line"].search([
```

ecino (changes requested):

ecino (approved):

PR #215, opened 2023-11-01
T0633 cancelled generated

jacrodile:


recurring_contract/models/contract_group.py
```diff
@@ -274,6 +274,16 @@ def _generate_invoices(self, invoicer):
                     })
 
                 else:
+                    # check if there is already an invoice for this partner and due date which has been cancelled
+                    all_invoices = self.env["account.move"].search([
+                        ("invoice_date_due", "=", invoicing_date),
+                        ("partner_id", "=", group.partner_id.id),
+                        ("state", "=", "cancel"),
+                        ("move_type", "=", "out_invoice")
+                    ])
```
ecino, line 283: This is a too broad check, beause he could have another invoice not related to the contract at all and this would prevent the invoice generation. You should add a filter similar to what is done for variable `already_paid_cl` at line 259.

ecino (changes requested):

PR #216, opened 2023-11-08
[T0657] Remove button from settings - issue fixed for all files

jacrodile:
No need to remove the button, I could fix it directly


NoeBerdoz:
@ecino This can be merged, the code updated fix the issue that let the code return true with bool() even when the string was "False",

ecino (approved):

PR #226, opened 2024-06-13
14 off balance

davidwul:
allow bypass lock dates for "off balance" (9xxxx) accounts


account_offbalance_sponsorship/security/ir.model.access.csv
```diff
@@ -0,0 +1,2 @@
+id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
+access_account_offbalance_sponsorship_account_offbalance_sponsorship,account_offbalance_sponsorship.account_offbalance_sponsorship,model_account_offbalance_sponsorship_account_offbalance_sponsorship,base.group_user,1,1,1,1
```
ecino, line 2: What's the purpose of this rule? The model `model_account_offbalance_sponsorship_account_offbalance_sponsorship` does not exist, and therefore the module update will fail with this.

ecino (changes requested):

ecino:
Still needs fixes.

PR #227, opened 2024-06-19
Fix t1438 list of invoices last payment date is wrong

minedcp:
fix last payment update when unreconciling


ecino (approved):
Is there a script to run in order to fix existing wrong data?

minedcp:
UPDATE account_move
SET last_payment = NULL
WHERE payment_state = 'not_paid';

PR #230, opened 2024-06-27
made sure unique invoices are being generated with correct dates

minedcp:
corected duplicated invoices as well as date of invoices generated incorectly

### DIFF
--- a/recurring_contract/models/contract_group.py
+++ b/recurring_contract/models/contract_group.py
@@ -229,7 +229,6 @@ class ContractGroup(models.Model):
         Context value async_mode set to False can force to perform
         the task immediately.
         """
-        self.ensure_one()
         invoicer = self.env["recurring.invoicer"].create({})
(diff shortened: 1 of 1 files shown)


recurring_contract/models/contract_group.py
```diff
@@ -229,6 +229,7 @@ def generate_invoices(self):
         Context value async_mode set to False can force to perform
         the task immediately.
         """
+        self.ensure_one()
```
ecino: This should not be here, the function handles recordset because of the `for group in self` loop.
```suggestion
```

ecino (changes requested):

recurring_contract/models/contract_group.py
minedcp: yes, i forgot to remove it. thanks

ecino (approved):

PR #239, opened 2024-08-15
T1657 - Fix invoice generation for Sub Proposal (1/3)

Prazn:
# Description
There is a bug that occurs when generating the invoices for a new Sub Proposal, some invoices are not generated at all.
When looking into it it seems that the check of if we need to generate the invoice was wrong because it did not take into account for which contract we are generating the invoice for. Meaning that if any other active contract already had an invoice for that month, we would skip it.

# Technical Aspects
There area lot of small changes here and there, to make things simple two aspects were changed :
- When generating invoices, if a specific contract is given we use it as a reference to check and generate invoices
- Fixed some conditions for when new invoice lines should be generated or "appended" to existing open invoices

# Misc
- Changes in `move.py` is just related to pre-commit formatting

### DIFF
--- a/recurring_contract/models/contract_group.py
+++ b/recurring_contract/models/contract_group.py
@@ -482,7 +482,7 @@ class ContractGroup(models.Model):
                     (
                         "product_id",
                         "in",
-                        contracts.product_ids.ids,
+                        contracts.mapped("product_ids").ids,
                     ),
                     ("payment_state", "=", "paid"),
                 ]
@@ -491,7 +491,7 @@ class ContractGroup(models.Model):
         )
         reference_contract = contracts[0]
         company_id = reference_contract.company_id.id
-        partner_id = self._get_partner_for_contract(contract).id
+        partner_id = self._get_partner_for_contract(reference_contract).id
         journal = self.env["account.journal"].search(
             [("type", "=", "sale"), ("company_id", "=", company_id)], limit=1
         )
--- a/recurring_contract/models/contract_group.py
+++ b/recurring_contract/models/contract_group.py
@@ -316,10 +316,9 @@ class ContractGroup(models.Model):
         - There is already an invoice for this due date which has been cancelled or
           edited.
         - Contract group suspension.
+        - A specific contract is given and an invoice for this due date have already exists and isn't cancelled
         """
         self.ensure_one()
-
-        search_filter = []
 
         if contract:
             # I changed the logic here... If the given contract already has an invoice for that month that is not cancelled -> skip it
@@ -373,10 +372,9 @@ class ContractGroup(models.Model):
 
     def _process_invoice_generation(self, invoicer, invoicing_date, contract=None):
         self.ensure_one()
-        # active_contracts = contract if contract else self.active_contract_ids
         active_contracts = self.active_contract_ids
         open_invoices = self.active_contract_ids.mapped("open_invoice_ids").filtered(
-            lambda i: i.invoice_date_due >= invoicing_date and i.invoice_date_due.year == invoicing_date.year # ADDED CHECK FOR SAME INVOICING YEAR
+            lambda i: i.invoice_date_due >= invoicing_date and i.invoice_date_due.year == invoicing_date.year
         )
 
         # invoice already open we complete the move lines
@@ -386,9 +384,9 @@ class ContractGroup(models.Model):
         open_invoice = open_invoices.filtered(
             lambda m: getattr(m.invoice_date_due, self.recurring_unit) == current_rec_unit_date
         )
-        if len(open_invoice) > 1: # NOT SURE ABOUT THIS, IT DOESN'T SEEM TO MAKE SENSE TO HAVE MORE THAN ONE
-            _logger.warning(
-                f"Found more than one open invoice on {invoicing_date} for {self.id}"
+        if len(open_invoice) > 1:
+            _logger.error(
+                f"Found more than one open invoice on {invoicing_date} for the group {self.id}"
             )
             return False
 
@@ -443,7 +441,12 @@ class ContractGroup(models.Model):
             )
         else:
             # Building invoices data
-            inv_data = self._build_invoice_gen_data(invoicing_date, invoicer, contract if contract is not None else self.active_contract_ids[0])
+            if contract is None:
+                # we use the first contract because the information we retrieve has to be shared
+                # between all the contracts of the list
+                contract = self.active_contract_ids[0]
+
+            inv_data = self._build_invoice_gen_data(invoicing_date, invoicer, contract)
             # Creating the actual invoice
             _logger.info(f"Generating invoice : {inv_data}")
             invoice = self.env["account.move"].create(inv_data)
@@ -480,8 +483,6 @@ class ContractGroup(models.Model):
             )
             .mapped("contract_id.contract_line_ids")
         )
-        # we use the first contract because the information we retrieve has to be shared
-        # between all the contracts of the list
         company_id = contract.company_id.id
         partner_id = self._get_partner_for_contract(contract).id
         journal = self.env["account.journal"].search(
@@ -521,7 +522,10 @@ class ContractGroup(models.Model):
                         invoicing_date=invoicing_date, contract_line=cl
                     ),
                 )
-                for cl in ( # not sure about this one bro... this is why the invoices are doubled/trippled/quadruppled ect...
+                # This loop is the cause in invoices being doubled/tripled/...
+                # We should only get here if there is not an existing open invoice for this period,
+                # so before there was no reason to reference the contract_line_ids of the other active invoices
+                for cl in (
                     contract.contract_line_ids - already_paid_cl
                 )
                 if cl
--- a/recurring_contract/models/contract_group.py
+++ b/recurring_contract/models/contract_group.py
@@ -269,7 +269,9 @@ class ContractGroup(models.Model):
                 )
 
                 # Check if invoice generation should be skipped for this date
-                if group._should_skip_invoice_generation(current_invoicing_date, contract):
+                if group._should_skip_invoice_generation(
+                    current_invoicing_date, contract
+                ):
                     continue
 
                 # Create a unique key for the invoice to track it
@@ -278,7 +280,9 @@ class ContractGroup(models.Model):
                 # Check if the invoice for this key has already been processed
                 if invoice_key not in processed_invoices:
                     # Process invoice generation if not already processed
-                    group._process_invoice_generation(invoicer, current_invoicing_date, contract)
+                    group._process_invoice_generation(
+                        invoicer, current_invoicing_date, contract
+                    )
                     # Add the invoice key to the set of processed invoices
                     processed_invoices.add(invoice_key)
 
@@ -336,7 +340,7 @@ class ContractGroup(models.Model):
                     "line_ids.product_id",
                     "in",
                     contract.product_ids.ids,
-                )
+                ),
             ]
         else:
             search_filter = [
@@ -356,7 +360,8 @@ class ContractGroup(models.Model):
                     "line_ids.product_id",
                     "in",
                     self.active_contract_ids.mapped("product_ids").ids,
-                )]
+                ),
+            ]
 
         existing_invoices = self.env["account.move"].search(search_filter)
 
@@ -376,7 +381,8 @@ class ContractGroup(models.Model):
         self.ensure_one()
         active_contracts = self.active_contract_ids
         open_invoices = self.active_contract_ids.mapped("open_invoice_ids").filtered(
-            lambda i: i.invoice_date_due >= invoicing_date and i.invoice_date_due.year == invoicing_date.year
+            lambda i: i.invoice_date_due >= invoicing_date
+            and i.invoice_date_due.year == invoicing_date.year
         )
 
         # invoice already open we complete the move lines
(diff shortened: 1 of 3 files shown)


recurring_contract/models/contract_group.py
```diff
@@ -485,9 +524,11 @@ def _build_invoice_gen_data(self, invoicing_date, invoicer, gift_wizard=False):
                         invoicing_date=invoicing_date, contract_line=cl
                     ),
                 )
+                # This loop is the cause in invoices being doubled/tripled/...
+                # We should only get here if there is not an existing open invoice for this period,
+                # so before there was no reason to reference the contract_line_ids of the other active invoices
                 for cl in (
```
Prazn: @ecino I think this is caused the issue  https://github.com/CompassionCH/compassion-accounting/pull/233 of multiple invoices for the same date that were generated.

This code is only executed if we have no open invoice for that month, but since it then looped on `active_contract_ids.contract_line_ids`, it would generate an invoice for every active contract, so 2 if there are 2 contract, 3 if 3...

recurring_contract/models/contract_group.py
```diff
@@ -304,57 +311,87 @@ def _calculate_start_date_and_offset(self):
...
         # invoice already open we complete the move lines
         current_rec_unit_date = getattr(invoicing_date, self.recurring_unit)
         # Keep invoice from the same month or year
         # (depending on the recurring unit)
         open_invoice = open_invoices.filtered(
-            lambda m: getattr(m.invoice_date_due, self.recurring_unit)
-            == current_rec_unit_date
+            lambda m: getattr(m.invoice_date_due, self.recurring_unit) == current_rec_unit_date
         )
+        if len(open_invoice) > 1:
```
Prazn, line 382: @ecino I'm not completely sure if we need this check but it seems weird to have multiple open invoices for the same invoice date. I had this locally, I don't know if it is the case in production.

Plus if there are multiple open invoices, it will "append" to those invoices a new line for this contract. So on the same invoice month the sponsor will pay multiple times for the same contract which seems wrong.

recurring_contract/models/contract_group.py
```diff
@@ -304,57 +311,87 @@ def _calculate_start_date_and_offset(self):
             offset = 1
         return start_date, offset
 
-    def _should_skip_invoice_generation(self, invoicing_date):
+    def _should_skip_invoice_generation(self, invoicing_date, contract=None):
```
Prazn, line 318: @ecino Can you take the time to see if the changes here are correct ? From what I've tested it seems to work as expected but I might have missed some use cases.

I tried to not have too many side effects, so limit my changes as much as possible. This is why I check specifically for a contract from a sub proposal, which might be overkill.

recurring_contract/models/contract_group.py
```diff
@@ -304,47 +315,76 @@ def _calculate_start_date_and_offset(self):
...
-            [
-                "|",
+
+        if contract:
+            search_filter = [
+                "&",
                 "&",
+                "&",
+                "&",
+                "&",
```
ecino: "&" is the default operator, so if there are only such comparisons, we can omit them and it's in my opinion more readable.
```suggestion
```

recurring_contract/models/contract_group.py
```diff
@@ -304,47 +315,76 @@ def _calculate_start_date_and_offset(self):
...
+            ]
+        else:
+            search_filter = [
                 "&",
                 "&",
                 "&",
                 "&",
-                ("move_type", "=", "out_invoice"),
-                ("state", "=", "cancel"),
+                "&",
```
ecino: I think again that you can remove all "&" operators here.

recurring_contract/models/contract_group.py
```diff
@@ -304,47 +315,76 @@ def _calculate_start_date_and_offset(self):
...
+                "|",
                 ("payment_state", "not in", ["paid", "not_paid"]),
+                ("state", "=", "cancel"),
```
ecino: Maybe those lines can be moved at the end of the domain to make a clearer separation and evident that the "or" applies to those two lines.

recurring_contract/models/contract_group.py
```diff
@@ -304,47 +315,76 @@ def _calculate_start_date_and_offset(self):
...
-        # Check for contract group suspension
-        is_suspended = (
-            self.invoice_suspended_until
-            and self.invoice_suspended_until > invoicing_date
-        )
 
-        return bool(dangling_invoices) or is_suspended
+        existing_invoices = self.env["account.move"].search(search_filter)
+
+        is_sub_proposal = contract is not None and contract.source_id == 554
```
ecino: I think this check is wrong because it's not good to hardcode an id in the code. It won't work for Nordic instances. Furthermore, it's better that all cases also check for group suspension in my opinion.

recurring_contract/models/contract_group.py
```diff
@@ -304,47 +315,76 @@ def _calculate_start_date_and_offset(self):
...
             ]
-        )
-        # Check for contract group suspension
-        is_suspended = (
-            self.invoice_suspended_until
-            and self.invoice_suspended_until > invoicing_date
-        )
 
-        return bool(dangling_invoices) or is_suspended
+        existing_invoices = self.env["account.move"].search(search_filter)
```
ecino: `search_count` is faster and sufficient in this case.
```suggestion
        existing_invoices = self.env["account.move"].search_count(search_filter)
```

recurring_contract/models/contract_group.py
```diff
@@ -355,6 +395,13 @@ def _process_invoice_generation(self, invoicer, invoicing_date):
             lambda m: getattr(m.invoice_date_due, self.recurring_unit)
             == current_rec_unit_date
         )
+        if len(open_invoice) > 1:
+            _logger.error(
+                f"Found more than one open invoice on {invoicing_date}"
+                f"for the group {self.id}"
+            )
+            return False
```
ecino: It's maybe a safety check but in that case shouldn't we simply skip to the regular invoice creation rather than aborting completely the invoice generation?
```suggestion
            _logger.error(
                f"Found more than one open invoice on {invoicing_date}"
                f"for the group {self.id}"
            )
            open_invoice = False
```

recurring_contract/models/contract_group.py
```diff
@@ -406,7 +453,12 @@ def _process_invoice_generation(self, invoicer, invoicing_date):
             )
         else:
             # Building invoices data
-            inv_data = self._build_invoice_gen_data(invoicing_date, invoicer)
+            if contract is None:
+                # we use the first contract because the information we retrieve
+                # has to be shared between all the contracts of the list
+                contract = self.active_contract_ids[0]
```
ecino: In my opinion this is wrong and would prevent to generate all lines of contracts in the regular case where we want to generate all invoice lines for a group. Only the first contract of the group will be generated and the others will miss. The method `_build_invoice_gen_data` should handle multiple contracts.

recurring_contract/models/contract_group.py
```diff
@@ -420,7 +472,9 @@ def _process_invoice_generation(self, invoicer, invoicing_date):
                 )
                 invoice.unlink()
 
-    def _build_invoice_gen_data(self, invoicing_date, invoicer, gift_wizard=False):
+    def _build_invoice_gen_data(
+        self, invoicing_date, invoicer, contract, gift_wizard=False
```
ecino:
```suggestion
        self, invoicing_date, invoicer, contracts, gift_wizard=False
```

recurring_contract/models/contract_group.py
```diff
@@ -485,10 +536,7 @@ def _build_invoice_gen_data(self, invoicing_date, invoicer, gift_wizard=False):
                         invoicing_date=invoicing_date, contract_line=cl
                     ),
                 )
-                for cl in (
-                    self.mapped("active_contract_ids.contract_line_ids")
-                    - already_paid_cl
-                )
+                for cl in (contract.contract_line_ids - already_paid_cl)
```
ecino:
```suggestion
                for cl in (contracts.mapped("contract_line_ids") - already_paid_cl)
```

ecino (changes requested):

recurring_contract/models/contract_group.py
Prazn: Done.

recurring_contract/models/contract_group.py
Prazn: Done.

recurring_contract/models/contract_group.py
Prazn: Done.

recurring_contract/models/contract_group.py
Prazn: Done.

recurring_contract/models/contract_group.py
Prazn: That part I really wasn't convinced when I made it...  my issue here was that we need to generate the "already paid" invoices (that come from the previous sponsorship) as well as the future invoices. But because of the group suspension these "already paid" invoices were never generated, only the future invoices did.

I'm not really sure how else to handle this case, this is why I have two different conditions.

recurring_contract/models/contract_group.py
Prazn: You are right, I changed _build_invoice_gen_data to take in a list of contracts, and generate the invoices for those contracts. So it should now handle both a list generation with a single contract and a list of multiple contracts.

I tested creating a new contract for a sponsor with 0 active contracts, it generated all the invoices correctly. And then for that same sponsor, created a new contract, and the invoices were correctly appended to the existing active invoices.

I'll let you see if the changes seems correct to you.

recurring_contract/models/contract_group.py
Prazn: Changed `_build_invoice_gen_data` to receive multiple contracts.

recurring_contract/models/contract_group.py
Prazn: Just curious, what is the different between using `contracts.contract_line_ids` or `contracts.mapped("contract_line_ids")` ? both seem to work

recurring_contract/models/contract_group.py
Prazn: If we are generating invoices for a specific contract it will work since we don't care about other contracts.

But in the case where we are handling a group I think we are going to create dupplicate invoices because we iterate over the active contract lines, then next time it's called tripple it, ect... I wanted to play it safe and stop the execution so that when it occurs, we can investigate it without creating wrong invoices.

recurring_contract/models/contract_group.py
ecino: It works when `contracts` is only one record. But the first solution will fail when multiple contracts are in the set.

recurring_contract/models/contract_group.py
```diff
@@ -432,21 +478,19 @@ def _build_invoice_gen_data(self, invoicing_date, invoicer, gift_wizard=False):
             .search(
                 [
                     ("date", "=", invoicing_date),
-                    ("contract_id", "in", self.active_contract_ids.ids),
+                    ("contract_id", "in", contracts.ids),
                     (
                         "product_id",
                         "in",
-                        self.active_contract_ids.mapped("product_ids").ids,
+                        contracts.product_ids.ids,
```
ecino:
```suggestion
                        contracts.mapped("product_ids").ids,
```

recurring_contract/models/contract_group.py
```diff
@@ -304,47 +315,66 @@ def _calculate_start_date_and_offset(self):
...
+        # Check for contract group suspension when no specific contract is given
+        # from a sub proposal is given
+        if is_sub_proposal:
+            return bool(existing_invoices)
+        else:
+            is_suspended = (
+                self.invoice_suspended_until
+                and self.invoice_suspended_until > invoicing_date
+            )
+            return bool(existing_invoices) or is_suspended
```
ecino: This check is not possible in this module. The method should be inherited in `sponsorship_compassion` and rather check on the field `parent_id` to determine if it's a SUB sponsorship and skip the suspension check in that case. Here, please keep only the regular check for every contract.

ecino (changes requested):

ecino:
You also have conflicts with the latest 14.0 version.

recurring_contract/models/contract_group.py
Prazn: Okay thanks, I adapted the code.

recurring_contract/models/contract_group.py
Prazn: Done.

recurring_contract/models/contract_group.py
Prazn: found a better way to handle different cases

recurring_contract/models/contract_group.py
Prazn: We decided to raise en error so that we can handle that case whe it comes up.

recurring_contract/models/contract_group.py
Prazn: We decided to raise en error so that we can handle that case whe it comes up.

recurring_contract/models/contract_group.py
Prazn: Moved in `sponsorship_compassion` like discussed. We also use the `child_id` to determine it is a sub sponsorship.

ecino (approved):

PR #240, opened 2024-08-19
T1721 - Generate gift wizard uses the wrong partner

clementcharmillot:
Create the invoice for the `send_gifts_to` target, defaults to `partner_id` if not set.

**Note**: The first contract is used to get `send_gifts_to`, as the current code does for other fields.

CompassionCH/compassion-modules#1949

### DIFF
--- a/recurring_contract/models/contract_group.py
+++ b/recurring_contract/models/contract_group.py
@@ -447,11 +447,7 @@ class ContractGroup(models.Model):
         # between all the contracts of the list
         contract = self.active_contract_ids[0]
         company_id = contract.company_id.id
-        partner_id = (
-            self.partner_id.id
-            if not contract.send_gifts_to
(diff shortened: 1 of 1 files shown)


ecino (changes requested):
The problem is that the field `send_gifts_to` is declared in the module `sponsorship_compassion`, which is a child from this module. Therefore, this field is not accessible from the module `recurring_contract` which could be theoretically installed without `sponsorship_compassion`. The fix has to be done inside the module `sponsorship_compassion` or any of its children.

clementcharmillot:
I've edited this PR to create a trivial method to get the `partner_id`.
The PR CompassionCH/compassion-modules#1949 in `compassion-modules` overrides this method with custom logic for `send_gifts_to`.

ecino (approved):

PR #241, opened 2024-08-21
T1712 - Delay automatic sub activation (2/2)

Prazn:
Follow-up of https://github.com/CompassionCH/compassion-switzerland/pull/1646

### DIFF
--- a/recurring_contract/models/recurring_contract.py
+++ b/recurring_contract/models/recurring_contract.py
@@ -537,7 +537,7 @@ class RecurringContract(models.Model):
     def invoice_paid(self, invoice):
         """Activate contract if it is waiting for payment."""
         activate_contracts = self.filtered(lambda c: c.state == "waiting")
-        if activate_contracts:
+        if self._can_activate_contract(activate_contracts):
             activate_contracts.contract_active()
 
     ##########################################################################
@@ -662,3 +662,6 @@ class RecurringContract(models.Model):
                     data_invs[inv_name] = inv_data
         if data_invs:
             self.mapped("open_invoice_ids").update_open_invoices(data_invs)
+
+    def _can_activate_contract(self, contract: "RecurringContract"):
+        return bool(contract)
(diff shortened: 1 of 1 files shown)


recurring_contract/models/recurring_contract.py
```diff
@@ -662,3 +662,6 @@ def _updt_invoices_rc(self, vals):
                     data_invs[inv_name] = inv_data
         if data_invs:
             self.mapped("open_invoice_ids").update_open_invoices(data_invs)
+
+    def _can_activate_contract(self, contract: "RecurringContract"):
+        return bool(contract)
```
ecino: - Theoretically, several contracts could be passed to the method. 
- Not really fan of the type annotation, especially because it's found nowhere else in our code. We should discuss if we want to introduce this practice?
- Maybe the return should be more explicit (True if any records are passed)?
```suggestion
    def _activable_contracts(self):
        return self.filtered(lambda c: c.state == "waiting")
```

recurring_contract/models/recurring_contract.py
```diff
@@ -537,7 +537,7 @@ def invoice_unpaid(self, invoice):
     def invoice_paid(self, invoice):
         """Activate contract if it is waiting for payment."""
         activate_contracts = self.filtered(lambda c: c.state == "waiting")
-        if activate_contracts:
+        if self._can_activate_contract(activate_contracts):
```
ecino:
```python
activable_contracts = self._activable_contracts()
if activable_contracts:
    activable_contracts.contract_active()
```

ecino (changes requested):

recurring_contract/models/recurring_contract.py
Prazn: Done

recurring_contract/models/recurring_contract.py
Prazn: 1. I didn't think about that case, is now handled.

2. I added it here because it made it simpler to work with the argument. I'll admit that I prefer it, since it means that you don't need to know/figure out what argument is given, but since the rest of the code doesn't use it, it was weird to have it here. Anyways since the signature changed it is removed.

3. I think the return type is fine, maybe I could change the function to be called `_get_activatable_contracts()` to be even more explicit but I feel that it's clear enough like this.

ecino (approved):

PR #242, opened 2024-08-30
T1657 - Fix dupplicate invoice generation with CRON

Prazn:
# Description
It looks like with the changes we made in this [PR](https://github.com/CompassionCH/compassion-accounting/pull/239) we introduced a new bug. When the CRON to generate invoices runs, dupplicate invoices for an already paid invoice are now generated.

# Technical Aspects
The issue comes from the `_should_skip_invoice_generation` for the case when we don't pass a specific contract. In that case the check would not find `existing_invoices` which returned `False`, and new invoices would be generated.

We changed the logic a bit, so we fetch all open invoices for the due date of that group, and if all active contracts are covered by those invoices then we don't need to generate new ones.

### DIFF
--- a/recurring_contract/models/contract_group.py
+++ b/recurring_contract/models/contract_group.py
@@ -317,11 +317,12 @@ class ContractGroup(models.Model):
 
     def _should_skip_invoice_generation(self, invoicing_date, contract=None):
         """In such cases, we should skip the invoice generation:
-        - There is already an invoice for this due date which has been cancelled or
-          edited.
-        - Contract group suspension.
         - A specific contract is given and an invoice for this due date already exists
           and isn't cancelled.
+        - All active contracts already have an invoice for this due date that
+          isn't cancelled.
+        - Contract group suspension.
+
         """
         self.ensure_one()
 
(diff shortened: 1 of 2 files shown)


recurring_contract/models/contract_group.py
```diff
@@ -349,19 +354,19 @@ def _should_skip_invoice_generation(self, invoicing_date, contract=None):
...
-                "|",
-                ("payment_state", "not in", ["paid", "not_paid"]),
-                ("state", "=", "cancel"),
+                ('state', '!=', 'cancel')
             ]
 
-        existing_invoices = self.env["account.move"].search_count(search_filter)
+            open_invoices = self.env["account.move"].search(search_filter)
+
+            has_all_invoices = len(self.active_contract_ids - open_invoices.line_ids.contract_id) == 0
```
ecino:
```suggestion
            has_all_invoices = len(self.active_contract_ids - open_invoices.mapped("line_ids.contract_id")) == 0
```

Prazn: Done

ecino (approved):

PR #243, opened 2024-09-04
Fix t1710 display correspondent field in survival sponsorship

Gordack:
Add correspondent fields to view and model.


ecino:
All changes should be made in modules `sponsorship_compassion` or `survival_sponsorship_compassion`. The fields you add here are already declared in `sponsorship_compassion` and should not be present in this generic accounting module. The only needed fix is probably to change the `invisible` domain of the field in the view in order to make sure contracts of type `CSP` are included.

PR #251, opened 2025-02-04
[MIG] RR/17.0 account offbalance sponsorship

robinrolle:
Port account offbalance sponsorphip

- OCA list of changes have been applied
- Refactor reconcile & Unreconcile ( See account.move_line/_add_off_balance_lines & account_move /js_remove_outstanding_partial)

There is still the issue with the payment_state od the account_move but it's not due to this module

Tested locally on my machine, please perform another check


account_offbalance_sponsorship/models/account_move.py
```diff
@@ -0,0 +1,115 @@
...
+
+        # Unreconcile
+        res = super().js_remove_outstanding_partial(partial_id)
+
+        # Get off-balance accounts
+        off_rec, off_ass = self.line_ids.get_account_offbalance(self.company_id)
+        account_32110 = self.env["account.account"].search([
+            ("code", "=", "32110"),
+            ("company_id", "=", self.company_id.id),
+        ], limit=1).id
```
ecino: Was it like this in v14? I wonder if we should either:

1. Move the module in compassion-nordic, because it's specific to them and only meant to be used by them
2. Implement a configurable approach to find the desired account without hardcoding the account code (would it be either `account_offbalance_receivable` or `account_offbalance_asset` added in the settings)?

@davidwul @robinrolle What do you think?

account_offbalance_sponsorship/models/account_move.py
```diff
@@ -0,0 +1,115 @@
...
+
+        """
+        # Get the standard reconciliation plan
+        plan_list, all_amls = self._optimize_reconciliation_plan(reconciliation_plan)
+
+        for plan_node in plan_list:
+            amls = plan_node['amls']
+
+            # Check if customer accoutn is affected
+            if any(aml.account_id.code.startswith("91") for aml in amls):
```
ecino: Same remark here for hardcoded account codes.

account_offbalance_sponsorship/models/account_move.py
```diff
@@ -0,0 +1,115 @@
...
+        Creates offset balance account_move_lines
+
+        :param plan_node: Dict containing reconciliation plan data.
+        """
+        company = plan_node['amls'][0].move_id.company_id
+        #Retrieve accounts defines in the settings
+        account_offbalance_receivable, account_offbalance_asset = self.get_account_offbalance(company)
+
+        for aml in plan_node['amls']:
+            if aml.account_id.code.startswith("91"):
```
ecino: Same remark with hardcoded account codes.

account_offbalance_sponsorship/models/account_move.py
```diff
@@ -0,0 +1,115 @@
...
+                            "account_id": account_offbalance_asset,
+                            "name": "off-balance asset",
+                            "move_id": move.id,
+                            "partner_id": move.partner_id.id,
+                            "debit": amount,
+                            "credit": 0.0,
+                        },
+                        {
+                            "account_id": self.env["account.account"].search([
+                                ("code", "=", "32110"),
```
ecino: Same remark.

account_offbalance_sponsorship/models/res_config.py
```diff
@@ -0,0 +1,48 @@
...
+class OffBalanceAccountConfigSettings(models.TransientModel):
+    """
+    Add the possibility to define an off balance asset account
+    and off balance receivable account
+    """
+
+    _inherit = "res.config.settings"
+
+    account_offbalance_receivable = fields.Many2one("account.account", readonly=False)
+    account_offbalance_asset = fields.Many2one("account.account", readonly=False)
```
ecino, line 23: Could we simply add the `company_dependent=True` property and remove the `get_values` and `set_values` overrides? To be Tested
```suggestion
    account_offbalance_receivable = fields.Many2one("account.account", company_dependent=True)
    account_offbalance_asset = fields.Many2one("account.account", company_dependent=True)
```

robinrolle:
Changed the logic so it's dynamic and not hard coded anymore. 

Tested the changes for the settings, didn't work

ecino (approved):

PR #253, opened 2025-02-12
[T2166] Fix: Update unpaid paste invoices when changing payment method (14 )

robinrolle:
Updated update_open_invoices to modify both future and overdue unpaid invoices when changing the payment method in the contract.


ecino:
Not a good change as discussed with David.

PR #261, opened 2025-06-25
fix: possibility of modifying advance_billing_month for groups of con…

jordyBSK:
possibility of modifying advance_billing_month for groups of contracts with several contracts


ecino:
The bug was already resolved with some other changes.

PR #264, opened 2025-10-17
changes to add the product in off balance accounting

davidwul:
the product needs to be part of the off-balance adjustement


davidwul:
@ecino can you review that 
and we need to see how to fill the products where it's empty in off-balance move line

PR #276, opened 2026-06-24
T2860 - FIX recurring invoice generation for annual contracts

danpa32:
Two bugs in _generate_invoices / _should_skip_invoice_generation:

1. Loop range used `advance_billing_months + 1` as upper bound, causing an off-by-one that generated too many invoices. Fixed to `advance_billing_months + month_interval - 1` so the number of generated periods is consistent across all billing cycles:
   - monthly/advance=1 → 1 invoice, monthly/advance=12 → 12 invoices
   - annual/advance=12 → 2 invoices (current + next year)

2. Skip check did not exclude cancelled invoices, so a super-deleted (cancelled) invoice for a future period would permanently block regeneration of that period. Fixed by adding `("state", "not in", ["cancel"])` to the search domain.

### DIFF
--- a/recurring_contract/README.rst
+++ b/recurring_contract/README.rst
@@ -59,10 +59,10 @@ Authors
 Contributors
 ------------
 
--  Cyril Sester <cyril.sester@outlook.com>
--  Emanuel Cino <ecino@compassion.ch>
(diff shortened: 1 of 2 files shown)


ecino (approved):
Looks good, thank you!

PR #278, opened 2026-06-29
REFACTOR Make queue_job dependencies optional

ecino:
- NEW module queue_job_sh_safe that will safely use queue_jobs when available and fallback to an alternate mechanism with ir_cron in case the module is not installed.
- This is useful for Nordic using odoo.sh on which queue jobs are prohibited and not optimal for the infrastructure


ecino:
LGTM

